PR #6, opened 2018-01-10
Latest TBC code

simoninns:
I've started the merge of the NTSC TBC code base with PAL - the configuration of the TBC is now controlled by a Tbc::setMode() function that allows you to ask for ntsc_cxadc (NTSC 28.8MSPS), ntsc_domdup (NTSC 32MSPS), pal_cxadc (PAL 28.8MSPS) and pal_domdup (PAL 32MSPS).  At the moment the ntsc_cxadc is the only working mode.  For the pal_domdup I've pulled in as many of the constants from the previous PAL TBC as possible.

You can also run the 'legacy' PAL TBC (32M) by specifying a command line switch which causes the code to use the other class.  (note that the command line switches have been cleaned up... a ./tbc -h will give you a full list of available switches and their meaning).

Basically there are a whole bunch of constants in the code with no description or clue what they mean.  In tbc.cpp I've marked them all with a comment beginning "// TODO" so you can find them all.  

In order to proceed I need you to run through the source code and comment on each of the constants.  Once I know what they are (and what's different between PAL and NTSC) I can update the code to make it react correctly.

As part of the code refactoring I've switched from the use of large arrays to C++ vectors.  Vectors allow run-time resizing of the storage buffers without the mess of direct memory allocation.  Functions (where possible) now automatically react to the size of a passed buffer (reducing the number of constants).  I've also removed all of the global buffers (and as many of the global variables as possible) - they are mainly passed-by-reference now.  For the globals that remain, I've grouped them into structs to make it clear what they are for and where they belong.


simoninns:
Just tested it with ld-decoder-ntsc32.py - it finds the frames and outputs valid data, but the burst detection doesn't function, so the output quality and colours are all over the place.  I'm guessing a tweak to burstDetect2 would cure most of it though.

Issue #155, opened 2019-02-19
Add cxadc LD capture sample support


happycube:
With oln's addition of 8-bit sample support and (early) VHS captures using same, it would make sense to (re-)add 8fsc/8bit capture support.  (this could potentially be done for rev5, but for now assign it to rev6)

simoninns:
This is supported in the current revision

PR #247, opened 2019-06-17
Add RF frequency option

atsampson:
Add an option to specify the RF sampling frequency, and fix a few places where it was assumed that `rf.freq` or `rf.freq_hz` was an integer.

Partly addresses #155. The remaining problem is that the ["repeat the above - but for [analogue] audio"](https://github.com/happycube/ld-decode/blob/61808aac41fd402e613d2e44086cdbdc9b8da568/lddecode_core.py#L497) code computes the array size wrongly at 8fSC, resulting in a "could not broadcast input array" exception.

With `-f 8fsc -p --daa`, this successfully decodes a PAL cx88 capture.

### DIFF
--- a/ld-decode.py
+++ b/ld-decode.py
@@ -34,6 +34,7 @@ parser.add_argument('--noEFM', dest='noefm', action='store_true', default=False,
 parser.add_argument('--daa', dest='daa', action='store_true', default=False, help='Disable analog audio decoding')
 parser.add_argument('--ignoreleadout', dest='ignoreleadout', action='store_true', default=False, help='continue decoding after lead-out seen')
 
+parser.add_argument('-f', '--frequency', dest='inputfreq', metavar='FREQ', type=parse_frequency, default=40, help='RF sampling frequency (default is 40MHz)')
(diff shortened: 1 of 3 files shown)


simoninns:
The reason for the delay on this is that we previously decided not to implement front-end configuration to allow use with sampling hardware other than the Domesday Duplicator (as this opens us up to a world of pain from users creating bug reports based on hardware that isn't supported, tested or owned by the developers).  We should discuss this pull request on the IRC before including.  (this is the same reason why the VHS work is currently forked too).

simoninns:
A thought on this - it might be good to split this; if we can include the code that allows ld-decode to use different sample rates, but not include the actual front end configuration switch; it would mean a simple patch if someone knows what they're doing and wants to play.  Could be a good compromise?

atsampson:
I've split this so adding the `-f` option is a separate commit at the end.

simoninns:
Great; I know it's a bit awkward, but I'm trying to keep us on track :)  Over to @happycube  for a final review then

PR #261, opened 2019-06-25
Performance improvements to ld-comb-pal

atsampson:
Profiling ld-comb-pal, I found a few things that could be tweaked to speed it up a bit: some changes to the palcolour inner loop, and some restructuring to the PalCombFilter thread pool to keep the worker threads busier.

Here are scaling boxplots (against ideal scaling curves) for the original version and the 1st, 4th and 5th commits below, on a quad-core i7-2600 machine built with `-O3 -march=native`. The Y axis is time in seconds for a 233-frame sample, and the X axis is the `-t` argument. Overall it's about 40% faster.

Figure_1

It may be possible to get a bit more speed out of it through some vectorisation; at present, GCC compiles nearly all of the inner loop as scalar instructions.

I've tried to follow the style of the existing ld-comb-pal/palcolour code. Let me know if there's anything that looks odd to you here...

### DIFF
--- a/tools/ld-comb-pal/filterthread.cpp
+++ b/tools/ld-comb-pal/filterthread.cpp
@@ -23,16 +23,15 @@
 ************************************************************************/
 
 #include "filterthread.h"
+#include "palcombfilter.h"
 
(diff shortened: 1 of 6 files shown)


simoninns:
Code revisions look amazing - many thanks!

PR #265, opened 2019-06-27
Add ldsflac script for compressing

Gamnn:
bash script for compressing .lds files with flac.


Gamnn:
using pv to pipe the lds into ld-lds-converter and silencing ffmpeg might be an improvement, to see the current progress. Would add a dependency, though.

Gamnn:
Going with a combined script.

Issue #251, opened 2019-06-18
Replace use of dddconv with local ld-decode-tools utility


simoninns:
Since local utilities are using dddconv, it would be a good idea to remove dddconv from the Domesday Duplicator repo and put it into ld-decode-tools instead.

Soon people without capture rigs will be decoding; so a cross-installation requirement is not the right way forward.  ld-cut.py will need updating too, which should be performed as part of this issue.

simoninns:
Added ld-lds-converter to the repo - basically this is the Domesday Duplicator's dddconv with the code modified to match the ld-decode-tools.  The command line switches are the same as dddconv.

Over to you @happycube to update ld-cut.py (and any other instance where you are making a cross repo reference to dddconv).

PR #271, opened 2019-06-28
Replace instances of dddconv to the in-repository ld-lds-converter tool

Gamnn:
Quick find & replace fix for #251


simoninns:
@happycube - can you check this quickly and merge?  I did a quick review and it looks fine.

Issue #263, opened 2019-06-27
Merge 2D NTSC comb filter code with ld-comb-pal


simoninns:
Since the 3D code is not functioning for NTSC and not present for PAL, it would be a good idea to merge the 2D comb-filters into a single code base (the 3D could then be added as a common function to both filters).  This would involve merging the 2D NTSC core filter into the PAL front-end (which is far more advanced than the current NTSC code and supports multithreading).

simoninns:
Related issues #211, #133, #126, #123, #119, #104, #93

atsampson:
This would also be a good opportunity for a more generic name -- ld-chroma-decoder? The current ld-comb-pal doesn't actually use a comb filter; it uses a pair of raised-cosine bandpass filters.

simoninns:
This is very true; the name 'comb-filter' was due to the previous implementation and that it is a commonly accepted name for a chroma decoder in the NTSC universe.  I like the idea of renaming it to ld-chroma-decoder.

simoninns:
Complete

PR #274, opened 2019-06-30
ld-comb-ntsc performance improvements

atsampson:
RFC - these changes make the single-threaded ld-comb-ntsc a bit over twice as fast, but some of this probably needs reworking before merging; let me know what you think.

As with ld-comb-pal, speeding up the filter code is the main contributor (35% speedup). I've added a new IIRFilter class that knows about the filter orders at compile time, which lets the compiler generate substantially better (specialised) code. If you're happy with this, I could adapt filtermaker and cx-expander to use it, and replace Filter entirely.

The filter code isn't vectorised, and it'd be difficult to do so without reworking how the YIQ data is stored; I'd suggest leaving this until after #263 since it'd be pretty similar to PALColour's filter core.

The other improvements are just fixing up things that stick out in the profiling results.

I've added a unit test for IIRFilter; is there a location or naming convention that you like for tests?


simoninns:
I'm happy with the changes; but @happycube should take a quick review of the filter side of things (as I think he's using that elsewhere in ld-decode?).  As for the unit tests, you are welcome to generate a directory for it; test or unittest as a directory name (etc) is fine.

PR #303, opened 2019-07-31
Use ffmpeg resampling to handle non-standard sample rates

atsampson:
This seems to work pretty well - I've tried it with a number of 8fSC captures and it produces better results than running the rest of the decoder at 8fSC. It should simplify future testing as well.

The script i used to test this with a few different input formats is [here](https://github.com/atsampson/ld-decode-testsuite/blob/master/try-ld-decode).

There's probably enough loader code in lddutils to make it worth moving into a module of its own now (lddecode.loaders?).


simoninns:
@happycube - can you verify this and pull it in?  Doesn't look like a change that is risky

PR #316, opened 2019-08-18
Strip out the parts of ld-ldstoefm that are no longer used.

atsampson:
ld-decode now does the filtering internally, so ld-ldstoefm is only doing the PLL.

You will need to rebuild ld-ldstoefm after this, because the arguments have changed.


happycube:
Thanks!

PR #319, opened 2019-08-25
3D Transform PAL support, and related features

atsampson:
A lot of refactoring, followed by some new features:

- `-f transform3d` is a 3D Transform PAL decoder, using the slower-but-better-quality of the two FFT structures that Jim Easterbrook's page recommends. The output is at least as good as transform2d on typical scenes, and better on anything with substantial amounts of cross-colour. How much slower it is is largely dependent on FFTW's performance on your machine (fancier processors will see much less difference between the two decoders).

- `--transform-mode`, or the checkbox in ld-analyse, lets you choose between the two different frequency-domain algorithms Jim describes. From initial testing, I think "threshold" (the one we had already) is better for 2D, and "level" is slightly better for 3D, but it's useful to be able to choose depending on the material, particularly for still images and noisy sources.

- `--show-ffts`, or another checkbox in ld-analyse, visualises the frequency-domain filtering in 2D and 3D mode as an overlay to the picture. The left column of boxes is the input to the filter, and the right column is the output.

Using the same sample image from GGV1011 as last time, here's pal2d:

pr-motion123-output

Here's transform2d, with FFT visualisation enabled:

pr-motion123-t2d-output

And here's transform3d, which gets rid of the remaining cross-colour in the upper windows, and reduces the ringing off the edge of the roof:

pr-motion123-t3d-output

At present, ld-analyse can only use the 2D decoders. I've provided a helper function in the SourceField class that can be used to load an appropriate batch of fields to satisfy all the decoders; see [here](https://github.com/atsampson/ld-decode/commit/b0fef2e0bbb5b92a0d39ade96b6b2b16c7bdd812#diff-889b59a70f46458fbd9c75c4dc90328e) for a rough example of how it might be used in ld-analyse if you want to do 3D PAL and NTSC there.


atsampson:
Some additional food for thought from this:

- I'd really like to know why the Z axis reflection didn't show up in the place I was expecting it! (I've probably got something backwards in my explanation...)

- SourceField, which groups together a field's data and metadata, saved a lot of effort in ld-chroma-decoder and could be useful for other tools; it may be worth moving into library/tbc.

- It would be straightforward to add other frequency-domain algorithms along the lines of TransformPal2D/3D - you just need to write a different applyFilter. This could be used, for example, to do a frequency-domain 3D notch filter for NTSC. (Or maybe someone's got a better idea for NTSC.)

- You can specify which location in the frame gets FFT-visualised with fields in PalColour::Configuration - this could be another mouse selection mode in ld-analyse. (Seeing decoder artefacts in a frame? Click on the artefact and you can see what the decoder's doing with it...)

Issue #332, opened 2019-09-17
VBI frame numbers on A-ko


Gamnn:
There's an odd pattern to the frame numbers on the first A-ko movie, on both the original and re-release. It comes out something like this, both in ld-decode and ld-process-vbi:

```
file frame 18521 unknown
file frame 18522 CAV frame 16665
file frame 18523 CAV frame 16663
file frame 18524 CAV frame 14765
file frame 18525 CAV frame 14766
file frame 18526 unknown
file frame 18527 CAV frame 16665
file frame 18528 CAV frame 16663
file frame 18529 CAV frame 14769
file frame 18530 CAV frame 14770
file frame 18531 unknown
```

Scrolling through on a player shows the proper frame numbers, without the odd non-sequential numbers after each pulldown frame (unknown).

[sample](https://drive.google.com/open?id=1-LlbBP_fjlF3eLUHpp5LZpOxmz_XoGRD)

atsampson:
In the frames that report incorrect numbers, VBI line 17 has a mangled value, but line 18 is correct:

```
Info: decoded 17= 16335365 18= 16335365 to picno= 14205
Info: decoded 17= 16335366 18= 16335366 to picno= 14206
Info: decoded 17= 16343007 18= 16335367 to picno= 16645
Info: decoded 17= 16343005 18= 16335368 to picno= 16643
Info: decoded 17= 16335369 18= 16335369 to picno= 14209
Info: decoded 17= 16335376 18= 16335376 to picno= 14210
Info: decoded 17= 16343005 18= 16335377 to picno= 16643
Info: decoded 17= 16343007 18= 16335378 to picno= 16645
```

e.g. line 17 for picno 14212 has the nonsense value 0xf95fdf, rather than 0xf94212.

A reasonably simple fix in `vbidecoder.cpp` would be to check that the BCD digits are actually in the range 0-9 before trusting the value -- this would also help with dropouts in the VBI.

simoninns:
Good idea @atsampson - also the fallback for a failed VBI code could also be to try to use line 18 instead (since the check condition would be there anyway).

simoninns:
Fixed

Gamnn:
It's still a little off, but only in the VBI frame number ranges of 9000-9999, 19000-19999, and 28000-28999 where it reports them as early CLV timecodes.

```
file frame 35068 CAV frame 27996
file frame 35069 CAV frame 27997
file frame 35070 CAV frame 27998
[...]

PR #335, opened 2019-09-20
Don't reject high CAV frame numbers

atsampson:
The top bit of the first picture number digit is used to signal (the lack of) a picture stop on some discs, so it needs to be masked off before the BCD is decoded in the code added to ld-decode for #332. (tools/library was doing this already so no change is needed there.)

For example, with side 4 of [CC1390L](https://www.lddb.com/laserdisc/06049/CC1390L/This-Is-Spinal-Tap:-Special-Edition), before this fix:

```
file frame 50000 unknown
file frame 50001 unknown
file frame 50002 unknown
file frame 50003 unknown
file frame 50004 unknown
```

And after:

```
file frame 50000 CAV frame 41209
file frame 50001 unknown
file frame 50002 CAV frame 41210
file frame 50003 CAV frame 41211
file frame 50004 CAV frame 41212
```

### DIFF
--- a/lddecode_core.py
+++ b/lddecode_core.py
@@ -2359,6 +2359,9 @@ class LDdecode:
                 self.isCLV = True
                 #logging.info('CLV', mins)
             elif (l & 0xf00000) == 0xf00000: # CAV frame
+                # Ignore the top bit of the first digit, used for PSC
+                l &= 0x7ffff
+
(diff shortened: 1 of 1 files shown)


happycube:
duh, silly me.   thanks for fixing it :)

PR #356, opened 2019-11-30
Use PyFFTW's FFT implementation if available

atsampson:
I'm not really sure this is worth it, but I'll let @happycube decide!

PyFFTW provides an interface to FFTW with the same API as numpy.fft. It's about 5% faster than numpy 1.17 over a 100-frame sample.

I also tried it without PyFFTW's cache enabled (which was slower), and with a more complex version that used FFTW_PATIENT mode, saving the FFTW wisdom to a file (which was... exactly the same as the default FFTW_ESTIMATE mode).

Figure_1

### DIFF
--- a/lddecode_core.py
+++ b/lddecode_core.py
@@ -24,6 +24,16 @@ import scipy as sp
 import scipy.signal as sps
 import scipy.interpolate as spi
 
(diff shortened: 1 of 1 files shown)


happycube:
Since it doesn't change any code flow, why not?  Thanks!

PR #362, opened 2019-12-01
Update the C++ EFM PLL to match the Python version

atsampson:
Most of this is just refactoring. The only functional change is to remove the hysteresis (prevDirection) in the zero-crossing detector, which results in a slight improvement in accuracy.

I've checked that this produces identical output to the Python version. I haven't updated the copy in cd-decode since I don't have any CD samples to test against yet, but I expect you could just drop this straight in (i.e. change the .pro to use ../ld-ldstoefm/pll.cpp).


atsampson:
Scrapping this because I've now got a better version.

PR #396, opened 2020-01-02
Add unit tests for VbiDecoder

atsampson:
Add unit tests for VbiDecoder (which are now run by the CI), and fix some minor problems they revealed:

- Don't assume a disc is CAV until we've seen a code indicating that.
- Fix the programme status code parity check - I believe this works for Amendment 2 codes too.
- Check BCD digits are valid in all the codes that use BCD - I've added a helper function for this.
- Don't store invalid user codes.

I've added tests for several of the oddities we've seen on real discs, but if anyone's got any more it'd be easy enough to add them.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -37,6 +37,10 @@ jobs:
       timeout-minutes: 5
       run: tools/library/filter/testfilter/testfilter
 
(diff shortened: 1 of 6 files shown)


simoninns:
Gave the code a quick browse through; changes look like a nice improvement in readability and the like, thanks!

PR #399, opened 2020-01-04
Add ld-export-metadata

atsampson:
This is a new command-line tool for converting information from the JSON file into formats that other tools can understand. At present, it can do:

- CSV export of VITS and VBI information (as ld-analyse already does)
- FFMETADATA1 export of chapter information decoded from VBI, ported from my decode-vbi script; you can use this to automatically add chapters to Matroska files.

I've tested FFMETADATA1 export against all the decoded discs I had around, and it seems to produce reasonable results for anything that has chapters.

Exporting CC subtitles as SRT would be a useful future extension.


atsampson:
Review question: I put the FFMETADATA1 code in ld-export-metadata, because it's a bit too complex (and magical) to be an LdDecodeMetaData method. Should the other two existing export modes be the same? Or should they all be in, say, library/tbc/export.cpp...?

simoninns:
I'm not a big fan of duplicating functionality :)  I also think a separate tool is the better approach to metadata export.  I'd be happy if both export functions in the metadata library were removed and placed in the ld-export-metadata app locally.  Then both functions can be removed from the library and ld-analyse as they will not be required.

atsampson:
This now moves all the export code into ld-export-metadata, and removes the ld-analyse GUI bits for it.

PR #408, opened 2020-01-17
Fix index types in testfilter

atsampson:
@Gamnn reported that Clang (I tested with LLVM 8.0.1) gives a warning for the first of these, and a narrowing error for the second.

### DIFF
--- a/tools/library/filter/testfilter/testfilter.cpp
+++ b/tools/library/filter/testfilter/testfilter.cpp
@@ -155,7 +155,7 @@ void testFIRFilter(const string &name, const Input& input, const Output& output,
     SimpleFilter refFilter(coeffs, one);
 
     // SimpleFilter has a delay, so pre-feed an appropriate number of samples
-    const int delay = coeffs.size() / 2;
+    const unsigned delay = coeffs.size() / 2;
     for (unsigned i = 0; i < delay; i++) {
         const double input_o = (i >= input.size()) ? 0 : input[i];
         refFilter.feed(input_o);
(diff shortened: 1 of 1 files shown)


atsampson:
The problem here is that vector::size's return type will depend on the platform's pointer size, but the test code ends up computing a fixed-size sample value from the index, so you have to have a potentially-narrowing conversion somewhere in the middle. (And I used the wrong one!)

We don't normally see this problem with code using Qt's collections like QVector because their size() returns int on all platforms (which IIRC dates back to the early days of Qt 4). The downside of that design is that on modern platforms QVector's total size is [artifically limited to < 2GiB](https://doc.qt.io/qt-5/qvector.html#maximum-size-and-out-of-memory-conditions), which is a bit of a pain.

I'm generally a bit dubious about the benefits of using qint32 for array indexes in the code (as opposed to using fixed-size types for things that actually are fixed sizes, like sample values). Qt 5 only supports platforms with 32-bit int, and qint32 is in practice always a typedef for int at the moment. And in the unlikely case that a future version of Qt supported platforms with bigger int, we'd then see conversion problems like this in many of the places where we're using QByteArray::size, etc.

(This also made me look at how qreal is defined -- it's not actually intended as a fixed-size floating point type, it's the type used for coordinates within Qt GUIs (QT_COORD_TYPE), and you can override it to be a different type when configuring Qt. So it's the right thing to use if, say, you're computing the position of a slider, but we probably shouldn't be using it for actual maths!)

simoninns:
Then, no problems for me; thanks for the explanation :)

Gamnn:
Builds on clang now. It does still throw some warnings about missing braces, (deemp.h has a bunch of these too), but it builds. Thanks!

PR #409, opened 2020-01-18
Tweaks for Mac compilation.

Gamnn:
Allows building with homebrew-installed qwt, and works around QT bug on macOS 10.12 (and possibly others) where ld-analyse wouldn't refresh after pressing frame buttons.

### DIFF
--- a/tools/ld-analyse/dropoutanalysisdialog.h
+++ b/tools/ld-analyse/dropoutanalysisdialog.h
@@ -26,11 +26,20 @@
 #define DROPOUTANALYSISDIALOG_H
 
 #include <QDialog>
+/* Workaround to compile on Mac. Similar may need to be done for other non-ubuntu distros */
+#if defined(__APPLE__)
+#include <qwt_plot.h>
+#include <qwt_plot_canvas.h>
+#include <qwt_legend.h>
+#include <qwt_plot_grid.h>
+#include <qwt_plot_curve.h>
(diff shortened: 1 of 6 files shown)


atsampson:
Adding `-I/usr/include/qwt` then doing `#include <qwt_whatever.h>` is probably the neatest fix - that's what GNURadio does, for example. (While installing Qwt's headers into a subdirectory isn't the default in Qwt's build system, it's the usual practice in GNU/Linux distros to reduce /usr/include pollution - e.g. anything Debian- or CentOS-derived has that layout - so I'd be tempted to suggest that Homebrew should do the same.)

For `isnanf`, add `#include <cmath>` and change the code to use `isnan` instead? It has overloads for float and double in C++ (and in C these days).

It would be a good idea to use `git rebase -i` and `git push -f` to squash your updates into your original commit so you don't break bisection.

Gamnn:
Fixed as @atsampson suggested. I think the homebrew-specific stuff should be okay for now; the paths are no longer version-specific at least. So it shouldn't break. I'd like to keep the mac building instructions simple by just needing to run brew install, and not have to do manual linking.

As an aside, I think there'll still be building issues on some other non-Ubuntu distros, due to Ubuntu's qwt-qt5 naming on it instead of qwt, but I'm uncertain of an easy work-around for that. I've left a comment in the code as a hint though.

PR #413, opened 2020-01-19
Icons for tools

Gamnn:
This set for ld-analyse should match Ubuntu a little better. Color scheme matched to the humanity icon set. I think it'll do until one of us figures out a theme-based solution.

Additionally, mac application icons added for ld-analyse and ld-process-efm, and the .ai files are included in the Graphics folder of each.


Gamnn:
Giving it some thought, I think just using custom icons might be preferable to theme icons anyway. The previous frame/next frame buttons are better represented by step backward / step forward icons which doesn't seem to be in any icon set, and the oscilloscope button would likely be better represented by something other than a mouse (I went with crosshairs here, but a logo of scope screen showing a stair pattern might be more intuitive).

Issue #462, opened 2020-02-13
NTSC framing


Gamnn:
For most of my sources, it seems to set the start and end as follows:
```
  "activeVideoStart": 134,
  "activeVideoEnd": 887,
```
Which comes out to 753 width. This often cuts off the right side of the picture and pads it with black to meet 760 width. Ideally, it should set the activeVideoEnd to (activeVideoStart + 760), unless (activeVideoStart + 760) > fieldWidth

Gamnn:
Solved with 5f2eb48.

PR #464, opened 2020-02-13
crack at #462

Gamnn:
Output NTSC activeVideoEnd as 760 from activeVideoStart, if it doesn't exceed fieldWidth. There's probably a more elegant way to do this.

### DIFF
--- a/lddecode/core.py
+++ b/lddecode/core.py
@@ -75,7 +75,7 @@ SysParams_NTSC = {
     'audio_rfreq': (1000000*315/88/227.5) * 178.75,
 
     'colorBurstUS': (5.3, 7.8),
-    'activeVideoUS': (9.45, 63.555-1.5),
+    'activeVideoUS': (9.45, 63.555-1.0),
 
     # In NTSC framing, the distances between the first/last eq pulses and the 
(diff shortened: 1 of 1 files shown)


Gamnn:
This is a bit hack-y. I _think_ changing line 73 in core.py to
```
    'activeVideoUS': (9.45, 63.555-1.0),
```
and line 93 to
```
SysParams_NTSC['activeVideoUS'] = (9.45, SysParams_NTSC['line_period'] - 1.0)
```
might work, but I've not had time to test.

simoninns:
The timings should be according to the NTSC specifications as-is?  Either way, I'd advise a complete regression test of all tools for NTSC before this is merged as the timings are used by pretty much everything in the processing chain.

simoninns:
In other words @Gamn - did you test this change across all tools in your branch before requesting a merge?

Gamnn:
I've tested the hacky-solution that only modifies the json output, but not the change to activeVideoUS, which I haven't committed yet, pending input from @happycube.

Gamnn:
activeVideoUS adjustment tested and committed. Seems good on what I've tested so far.

PR #475, opened 2020-03-21
Use TBC library logging functions in ld-chroma-encoder

atsampson:
Same changes as the other main.cpp files.


atsampson:
I'll include this with a bigger pull request I'm about to submit instead.

Issue #478, opened 2020-04-05
Video LPF vs. a real player


atsampson:
I've been experimenting with decoding the PAL Lasers disc this weekend, and found it looks better if you apply some equalisation to the TBC in order to make the VITS signals look correct -- it needed both some HF gain and some phase correction. (In addition to luma sharpness, Transform PAL will work better the flatter the frequency response is.)

This made me wonder how the frequency response in the TBC compares to the composite output of a real player. So I've used the DDD to capture both RF and composite output from my calibrated LD-V4300D for a few clips - because it's not DC coupled the composite output is a mess, but you can see the important parts of the VITS.

Science Museum Lasers (PAL, CAV, start of the disc) - all the bursts should have the same amplitude:

![LasersBursts](https://user-images.githubusercontent.com/436317/78512418-3a017900-779c-11ea-91f5-7cd51b2eba07.png)

Jason and the Argonauts (PAL, CAV, end of the disc):

![JasonBursts](https://user-images.githubusercontent.com/436317/78512416-34a42e80-779c-11ea-8170-1dd88b09444d.png)

Video Essentials (NTSC, CAV, middle of the disc):

![VEBursts](https://user-images.githubusercontent.com/436317/78512419-3e2d9680-779c-11ea-9049-6fb979526cbb.png)

NTSC looks pretty good, but for PAL the 4.8 MHz burst is noticably attenuated in ld-decode's output relative to what the player produces...

The video LPF is an order 7 Butterworth filter with a cutoff of 4.8 MHz by default (actually applied in FFT form as usual). As an experiment, making it order 25 (!) at 5.2 MHz produces something pretty close to the real player:

![JasonBursts-5 2order25](https://user-images.githubusercontent.com/436317/78512531-42a67f00-779d-11ea-94a0-ab9dcfd25bf3.png)

But the high order means it's very non-phase-linear, which shows up in the VITS phase test - the two-frequency pulse in the middle should be symmetrical, with the bottom peaks flat at 0V:

![JasonPulse-5 2order25](https://user-images.githubusercontent.com/436317/78512547-64076b00-779d-11ea-8739-f1f95833c6fd.png)

Any ideas for a better filter profile, @happycube? (I wonder if directly generating the FFT form would work better...)

happycube:
Direct FFT filter generation would be interesting indeed.

Sounds like a higher LPF frequency is needed on PAL, possibly with a lower filter order.

atsampson:
[...]

PR #481, opened 2020-04-11
Remove burst-based chroma gain calculation

atsampson:
This is a "request for comments" PR rather than a "merge immediately" PR :-)

At the moment, the NTSC and PAL decoders adjust the gain they apply to the IQ or UV signals based on the `medianBurstIRE` value that ld-decode reports in the JSON. According to the comments, this was intended to act as a makeshift MTF compensator, before ld-decode was able to do that itself.

Now that ld-decode has working MTF compensation, I think this mechanism is probably more trouble than it's worth - the relationship between the burst size and the gain applied depends on a fudge factor (different for PAL/NTSC), and when feeding ld-chroma-decoder with known-correct data you have to provide a synthetic medianBurstIRE value and disable the fudge factor.

This set of changes stops ld-chroma-decoder from looking at `medianBurstIRE` entirely, so the chroma gain is just based on the black/white levels in the TBC (see comments in the first commit).  For troublesome discs, `--chroma-gain` now defaults to 1.0 and works for both PAL and NTSC, and ld-analyse gets a chroma gain slider.

(Somewhat related to #478 in that the current video LPF rolls off a bit more PAL chroma than it should.)

### DIFF
--- a/tools/ld-analyse/chromadecoderconfigdialog.cpp
+++ b/tools/ld-analyse/chromadecoderconfigdialog.cpp
@@ -32,6 +32,9 @@ ChromaDecoderConfigDialog::ChromaDecoderConfigDialog(QWidget *parent) :
     ui->setupUi(this);
     setWindowFlags(Qt::Window);
 
+    ui->chromaGainHorizontalSlider->setMinimum(0);
+    ui->chromaGainHorizontalSlider->setMaximum(200);
+
     ui->thresholdHorizontalSlider->setMinimum(0);
     ui->thresholdHorizontalSlider->setMaximum(100);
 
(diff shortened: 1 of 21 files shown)


atsampson:
ld-analyse question for @simoninns : the chroma gain slider would be useful for both NTSC and PAL. But should there be a separate "NTSC chroma decoder settings" dialog, or a shared dialog for both? (I'm leaning towards the latter since there aren't that many settings overall...)

simoninns:
I'm fine with the changes - I think, on the dialogue side of things, perhaps a tabbed dialogue is better with the PAL and NTSC settings in the same window on separate tabs.  Keeps the overall number of dialogues down, whilst still keeping the UI obvious and clean.

atsampson:
Now with fancier PAL/NTSC chroma decoder options window. The options not applicable to the current standard are greyed out as before, and the tab for the current standard is selected automatically when you open a file.

cdc-pal
cdc-ntsc

happycube:
As I said in #490, I think this should eventually be reinstated but in improved form, using VITS to guide it - there's usually a lot of info there that ld-decode's never used.  When MTF compensation isn't working correctly (or sometimes too far into a CAV disk for it to have any) the color burst is non-linear with luma.

PR #491, opened 2020-05-25
Add options for NTSC noise reduction levels

atsampson:
@Gamnn spotted that these were missing - you could adjust them in ld-analyse but not from the command line.


happycube:
<Gamn> speaking of merging in, I think the pull request on the ld-chroma-decoder options can be. I've only tested the luma option, but haven't seen any problems with it

PR #512, opened 2020-07-13
Add some explanation to the NTSC chroma filter code

atsampson:
I didn't entirely understand how this worked before, and there were a couple of comments saying "This could use some explanation"... so I've added some. No changes to behaviour.

Reviewer set to @happycube since you wrote this originally - please can you check that I've understood it correctly?


atsampson:
Updated after a bit more thought to add a note on why split2D uses split1D's output...

Issue #513, opened 2020-07-14
Improve Comb::filterIQ's filter


atsampson:
`Comb::filterIQ` is the IQ post-filter in the NTSC decoder - in a conventional decoder, it would filter off aliases in the I/Q signals after demodulation.

At present, it doesn't actually do anything because of a bug in `Comb::decodeFrame`: in both places where it's called, it's applied to `currentFrameBuffer.yiqBuffer` rather than `tempYiqBuffer`. If you fix that, it doesn't work very well (the filter delay compensation needs adjusting, and its filters aren't linear-time anyway).

The NTSC decoder seems to work perfectly well without the postfilter - I guess the demodulator effectively works like a [Tayloe detector](http://www.norcalqrp.org/files/Tayloe_mixer_x3a.pdf) rather than a conventional mixer. So should I fix the bug above but have it turned off by default, or should I just remove `filterIQ` entirely?

atsampson:
Looking at the old motion detection code, that had the same problem:

```
        splitIQ(&currentFrameBuffer);

        tempYiqBuffer = currentFrameBuffer.yiqBuffer;

        adjustY(&currentFrameBuffer, tempYiqBuffer);
        if (configuration.colorlpf) filterIQ(currentFrameBuffer.yiqBuffer); // <-- here
        doYNR(tempYiqBuffer);
        doCNR(tempYiqBuffer);

        opticalFlow.denseOpticalFlow(currentFrameBuffer.yiqBuffer, currentFrameBuffer.kValues); // <-- and here
```

`splitIQ` leaves the baseband signal in `y`, so `denseOpticalFlow` was seeing the baseband signal rather than Y. That may be another reason why it never worked very well...

simoninns:
I think this is a question for @happycube  - but my vote would be 'remove'

Gamnn:
It's functional in master now.

Off:
![frame_ntsc_comb_47505](https://user-images.githubusercontent.com/44366267/89704390-6e758380-d98e-11ea-9bb3-2193e4525a15.png)

On:
![frame_ntsc_comb_47505pf](https://user-images.githubusercontent.com/44366267/89704393-759c9180-d98e-11ea-8514-b1367c47dda8.png)

simoninns:
So, what are we actually saying here... the issue is solved and should be closed?

Gamnn:
Perhaps not entirely. It works in ld-analyse, but I don't see an option for it in ld-chroma-decoder.

[...]

PR #516, opened 2020-07-18
Enable 3D decoders in ld-analyse

atsampson:
~~This is on top of the changes in #512, so don't merge until that one's been approved - it may need a rebase.~~

The main thing here is reworking `Comb` so that it provides the same `decodeFrames` interface for 3D decoding as `PalColour`, then making ld-analyse use it. Because `Comb` now has access to the complete batch of fields being decoded it should make future NTSC 3D work a bit easier too.

There are now radio buttons in the chroma decoder dialog that allow you to select 3D decoding for NTSC and PAL. I've left the defaults as 2D, because NTSC 3D isn't generally useful and Transform PAL 3D is a bit slow for interactive use.

This also includes a fix for the `Comb::filterIQ` problem in #513, and adds a checkbox for `colorLpf` so you can experiment with it.

### DIFF
--- a/tools/ld-analyse/chromadecoderconfigdialog.cpp
+++ b/tools/ld-analyse/chromadecoderconfigdialog.cpp
@@ -4,6 +4,7 @@
 
     ld-analyse - TBC output analysis
     Copyright (C) 2019 Simon Inns
+    Copyright (C) 2020 Adam Sampson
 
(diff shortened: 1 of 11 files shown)


happycube:
I committed the other pull request as is, but I think @simoninns should look at this one (time permitting)

PR #522, opened 2020-08-03
New adaptive mode for ntsc3d decoder

atsampson:
3D decoding looks for candidates in the current frame and two fields in either direction, using a similarity heuristic based on the output of the existing 2D decoder. The heuristic is a quality tradeoff: it can produce worse results than the non-adaptive 3D decoder for absolutely still images, but overall the results are much better, particularly for animation and other material with repeated fields/frames.

This is a much-tidied-up version of the branch that @Gamnn has been testing, so there are a couple of known problems that can fool the 3D heuristic: especially bad results from the 2D decoder (e.g. dotcrawl.ldf, Video Essentials frame 29108), and low-saturation colours (e.g. scarf.ldf, VE frame 36430). Any future improvements to ntsc2d will generally benefit ntsc3d as well.

Most of the commits are refactoring Comb; the 3D stuff is all in the last commit. Performance-wise, ntsc2d is a few percent faster than it was before, and ntsc3d is pretty slow (~10 FPS on one core; it would benefit from some vectorisation).

I'll update the docs for ld-chroma-decoder and ld-analyse once this is merged.


atsampson:
@Gamnn pointed out that the adaptive checkbox didn't work - fixed.

happycube:
Indeed - approved and merged.  Thank you!

PR #616, opened 2021-02-06
tools: add YNR to PAL

happycube:
This adds support for luma noise reduction in PAL.  Set to 0.5 IRE at the moment to try to take some of the edge off the noise, but not remove too much detail.  On some disks this may need to be set to 2-3 IRE.

Includes ld-analyse and ld-chroma-decoder adjustments, including the GUI, so please look it over and see that I did it the right way :)


tools/ld-chroma-decoder/palcolour.cpp
```diff
@@ -372,6 +374,36 @@ void PalColour::detectBurst(LineInfo &line, const quint16 *inputData)
...
+    // High-pass filter for Y
+    auto yFilter(f_nr);
+
+    const double irescale = (videoParameters.white16bIre - videoParameters.black16bIre) / 100;
+
+    double nr_y = configuration.yNRLevel * irescale;
+
+    double hplinef[videoParameters.fieldWidth];
+
+    for (qint32 h = 0; h <= videoParameters.fieldWidth; h++) {
```
atsampson, line 389: < not <=, else it'll write off the end of the array. (The NTSC code has the same bug, looking at it.)

tools/ld-analyse/chromadecoderconfigdialog.cpp
```diff
@@ -44,6 +44,9 @@ ChromaDecoderConfigDialog::ChromaDecoderConfigDialog(QWidget *parent) :
 
     ui->yNRHorizontalSlider->setMinimum(0);
```
atsampson, line 45: How about having a single Luma NR slider at the top, under the chroma gain slider? (The Chroma NR slider would need to stay in the NTSC panel for now since there isn't an equivalent for PAL.)

tools/ld-analyse/chromadecoderconfigdialog.cpp
```diff
@@ -63,6 +66,7 @@ void ChromaDecoderConfigDialog::setConfiguration(bool _isSourcePal, const PalCol
 
     palConfiguration.chromaGain = qBound(0.0, palConfiguration.chromaGain, 2.0);
     palConfiguration.transformThreshold = qBound(0.0, palConfiguration.transformThreshold, 1.0);
+    palConfiguration.yNRLevel = qBound(0.0, palConfiguration.yNRLevel, 1500.0);
```
atsampson, line 69: Why 1500 - shouldn't that be 10 like below?

tools/ld-chroma-decoder/main.cpp
```diff
@@ -189,7 +189,7 @@ int main(int argc, char *argv[])
 
     // Option to set the luma noise reduction level
     QCommandLineOption lumaNROption(QStringList() << "luma-nr",
-                                    QCoreApplication::translate("main", "NTSC: Luma noise reduction level in dB (default 1.0)"),
+                                    QCoreApplication::translate("main", "Luma noise reduction level in dB (default 1.0)"),
```
atsampson, line 192: "(default 1.0 for NTSC, 0.5 for PAL)"?

tools/ld-chroma-decoder/palcolour.cpp
```diff
@@ -372,6 +374,36 @@ void PalColour::detectBurst(LineInfo &line, const quint16 *inputData)
     line.bq /= burstNorm;
 }
 
+// Perform analog-style noise coring.
+void PalColour::doYNR(double *Yline)
+{
+    // High-pass filter for Y
+    auto yFilter(f_nr);
```
atsampson, line 381: f_nr is designed for the NTSC sample rate and luma bandwidth. Would it be better to have two separate filters in filtermaker for this?

tools/ld-chroma-decoder/palcolour.cpp
```diff
@@ -519,7 +551,10 @@ void PalColour::decodeLine(const SourceField &inputField, const ChromaSample *ch
     // burst-based correction applied.
     const double scaledSaturation = 2.0 * scaledContrast * chromaGain;
 
-    for (qint32 i = videoParameters.activeVideoStart; i < videoParameters.activeVideoEnd; i++) {
+    // extract luma first so it can be run through NR
+    double extractedY[videoParameters.fieldWidth];
```
atsampson, line 555: I don't think it's worth changing this right now, since this loop doesn't vectorise very well at the moment anyway, but it might be worth putting this in an aligned array that's not on the stack in the future.

atsampson (commented):
Added a few comments - otherwise looks good!

simoninns:
Hi, I had a quick look through ld-analyse and tested the GUI changes - looks good to me,

happycube:
Thanks, both of you!

PR #618, opened 2021-02-13
Add BT.601 YCbCr output

ifb:
Add planar YCbCr output (YUV444P16 or GRAY16).

### DIFF
--- a/tools/ld-analyse/tbcsource.cpp
+++ b/tools/ld-analyse/tbcsource.cpp
@@ -536,7 +536,7 @@ QImage TbcSource::generateQImage(qint32 frameNumber)
         // Chroma decode the current frame and display
 
         // Decode colour for the current frame, to RGB 16-16-16 interlaced output
-        QVector<RGBFrame> outputFrames(1);
+        QVector<OutputFrame> outputFrames(1);
         if (videoParameters.isSourcePal) {
             // PAL source
             palColour.decodeFrames(inputFields, startIndex, endIndex, outputFrames);
@@ -546,7 +546,7 @@ QImage TbcSource::generateQImage(qint32 frameNumber)
         }
 
         // Get a pointer to the RGB data
-        const quint16 *rgbPointer = outputFrames[0].data();
+        const quint16 *rgbPointer = outputFrames[0].RGB.data();
 
         // Fill the QImage with black
         frameImage.fill(Qt::black);
--- a/tools/ld-chroma-decoder/comb.cpp
+++ b/tools/ld-chroma-decoder/comb.cpp
@@ -108,7 +108,7 @@ void Comb::updateConfiguration(const LdDecodeMetaData::VideoParameters &_videoPa
 }
 
 void Comb::decodeFrames(const QVector<SourceField> &inputFields, qint32 startIndex, qint32 endIndex,
-                        QVector<RGBFrame> &outputFrames)
+                        QVector<OutputFrame> &outputFrames)
 {
     assert(configurationSet);
     assert((outputFrames.size() * 2) == (endIndex - startIndex));
@@ -172,8 +172,9 @@ void Comb::decodeFrames(const QVector<SourceField> &inputFields, qint32 startInd
         currentFrameBuffer->doYNR();
         currentFrameBuffer->doCNR();
 
-        // Convert the YIQ result to RGB
-        outputFrames[frameIndex] = currentFrameBuffer->yiqToRgbFrame();
+        // Convert the YIQ result to RGB or YCbCr
+        outputFrames[frameIndex] = configuration.outputYCbCr ? currentFrameBuffer->yiqToYUVFrame() :
+                                                               currentFrameBuffer->yiqToRGBFrame();
 
         // Overlay the map if required
         if (configuration.dimensions == 3 && configuration.showMap) {
@@ -195,9 +196,9 @@ Comb::FrameBuffer::FrameBuffer(const LdDecodeMetaData::VideoParameters &videoPar
     irescale = (videoParameters.white16bIre - videoParameters.black16bIre) / 100;
 }
 
-/* 
+/*
  * The color burst frequency is 227.5 cycles per line, so it flips 180 degrees for each line.
- * 
+ *
  * The color burst *signal* is at 180 degrees, which is a greenish yellow.
  *
  * When SCH phase is 0 (properly aligned) the color burst is in phase with the leading edge of the HSYNC pulse.
@@ -211,7 +212,7 @@ Comb::FrameBuffer::FrameBuffer(const LdDecodeMetaData::VideoParameters &videoPar
 inline qint32 Comb::FrameBuffer::getFieldID(qint32 lineNumber) const
 {
     bool isFirstField = ((lineNumber % 2) == 0);
-    
+
     return isFirstField ? firstFieldPhaseID : secondFieldPhaseID;
 }
 
@@ -219,11 +220,11 @@ inline qint32 Comb::FrameBuffer::getFieldID(qint32 lineNumber) const
 inline bool Comb::FrameBuffer::getLinePhase(qint32 lineNumber) const
 {
     qint32 fieldID = getFieldID(lineNumber);
-    bool isPositivePhaseOnEvenLines = (fieldID == 1) || (fieldID == 4);    
+    bool isPositivePhaseOnEvenLines = (fieldID == 1) || (fieldID == 4);
 
     int fieldLine = (lineNumber / 2);
     bool isEvenLine = (fieldLine % 2) == 0;
-    
+
     return isEvenLine ? isPositivePhaseOnEvenLines : !isPositivePhaseOnEvenLines;
 }
 
@@ -662,14 +663,14 @@ void Comb::FrameBuffer::doYNR()
     }
 }
 
-// Convert buffer from YIQ to RGB 16-16-16
-RGBFrame Comb::FrameBuffer::yiqToRgbFrame()
+// Convert buffer from YIQ to RGB and store as packed RGB48
+OutputFrame Comb::FrameBuffer::yiqToRGBFrame()
 {
-    RGBFrame rgbOutputFrame;
-    rgbOutputFrame.resize(videoParameters.fieldWidth * frameHeight * 3); // for RGB 16-16-16
+    OutputFrame outputFrame;
+    outputFrame.RGB.resize(videoParameters.fieldWidth * frameHeight * 3); // for RGB 16-16-16
 
     // Initialise the output frame
-    rgbOutputFrame.fill(0);
+    outputFrame.RGB.fill(0);
 
     // Initialise YIQ to RGB converter
     RGB rgb(videoParameters.white16bIre, videoParameters.black16bIre, configuration.whitePoint75, configuration.chromaGain);
@@ -677,7 +678,7 @@ RGBFrame Comb::FrameBuffer::yiqToRgbFrame()
     // Perform YIQ to RGB conversion
     for (qint32 lineNumber = videoParameters.firstActiveFrameLine; lineNumber < videoParameters.lastActiveFrameLine; lineNumber++) {
         // Get a pointer to the line
-        quint16 *linePointer = rgbOutputFrame.data() + (videoParameters.fieldWidth * 3 * lineNumber);
+        quint16 *linePointer = outputFrame.RGB.data() + (videoParameters.fieldWidth * 3 * lineNumber);
 
         // Offset the output by the activeVideoStart to keep the output frame
         // in the same x position as the input video frame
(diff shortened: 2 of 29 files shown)


tools/ld-chroma-decoder/monodecoder.cpp
```diff
@@ -28,6 +28,12 @@
 #include "decoderpool.h"
 #include "palcolour.h"
 
+MonoDecoder::MonoDecoder(const Comb::Configuration &combConfig)
```
atsampson, line 31: This doesn't look right - each Decoder subclass has its own Configuration subclass; it doesn't really make sense to pass the NTSC decoder's configuration to MonoDecoder. As MonoDecoder doesn't need anything that isn't in Decoder::Configuration then it just inherits it directly; you should pass a MonoDecoder::Configuration here, not a Comb::Configuration (and then the constructor doesn't need to do anything to convert it).

tools/ld-chroma-decoder/decoder.h
```diff
@@ -83,14 +83,16 @@ class Decoder {
         LdDecodeMetaData::VideoParameters videoParameters;
         qint32 topPadLines;
         qint32 bottomPadLines;
-    };
+        QString pix_fmt = "RGB48";
```
atsampson, line 86: pixelFormat, not pix_fmt.

It would be better to make this an enumerated type, like the other choose-one-of-several options in the Configuration structs, not a string - QString does all sorts of horrid implicit-sharing stuff internally, and comparing QStrings later on isn't going to be very efficient (plus the compiler can't warn you if you forget to handle one of the possibilities).

tools/ld-chroma-decoder/main.cpp
```diff
@@ -152,6 +152,16 @@ int main(int argc, char *argv[])
                                         QCoreApplication::translate("main", "number"));
     parser.addOption(chromaGainOption);
 
+    // Option to output packed RGB (--rgb)
```
atsampson, line 155: It'd be simpler to just have one option -- either leave RGB as the default and have a --yuv flag to enable YUV output, or add an output format option that takes a string (like -f).

(If you have two boolean options, then you also need to check the user hasn't specified both of them below.)

tools/ld-chroma-decoder/main.cpp
```diff
@@ -229,7 +239,7 @@ int main(int argc, char *argv[])
     parser.addPositionalArgument("input", QCoreApplication::translate("main", "Specify input TBC file (- for piped input)"));
 
     // Positional argument to specify output video file
-    parser.addPositionalArgument("output", QCoreApplication::translate("main", "Specify output RGB file (omit or - for piped output)"));
+    parser.addPositionalArgument("output", QCoreApplication::translate("main", "Specify output file (omit or - for piped output). RGB48, YUV444P16, GRAY16 pixel formats are supported."));
```
atsampson, line 242: If you had an output-format option, the list of formats here could go into the help for that.

tools/ld-chroma-decoder/main.cpp
```diff
@@ -376,6 +396,11 @@ int main(int argc, char *argv[])
 
     if (parser.isSet(showFFTsOption)) {
         palConfig.showFFTs = true;
+        if (palConfig.outputYCbCr ) {
+            // Quit with error
+            qCritical("YUV output not available when showFFT is enable");
```
atsampson, line 401: enable*d*.

tools/ld-chroma-decoder/main.cpp
```diff
@@ -376,6 +396,11 @@ int main(int argc, char *argv[])
 
     if (parser.isSet(showFFTsOption)) {
         palConfig.showFFTs = true;
+        if (palConfig.outputYCbCr ) {
```
atsampson, line 399: Stray space before the ).

tools/ld-chroma-decoder/palcolour.cpp
```diff
@@ -574,29 +583,65 @@ void PalColour::decodeLine(const SourceField &inputField, const ChromaSample *ch
...
+
+        const double yScale = 219.0 * 257.0 / (videoParameters.white16bIre - videoParameters.black16bIre);
+        const double cbScale = 112.0 / 256.0 / (ONE_MINUS_Kb * kB); // Poynton, Eq 25.5 & 28.1
+        const double crScale = 112.0 / 256.0 / (ONE_MINUS_Kr * kR);
+
+        for (qint32 i = videoParameters.activeVideoStart; i < videoParameters.activeVideoEnd; i++) {
+            double rY = extractedY[i];
+
+            // Scale to 16-bit output
+            rY = qBound(1.0*256.0, (rY - videoParameters.black16bIre) * yScale + 16*256, 254.75*256.0);
```
atsampson, line 599: Spaces around *. Ditto below.

tools/ld-chroma-decoder/ycbcr.cpp
```diff
@@ -0,0 +1,98 @@
...
+void YCbCr::convertLine(const YIQ *begin, const YIQ *end, quint16 *outY, quint16 *outCb, quint16 *outCr)
+{
+    // Factors to scale Y according to the black to white interval
+    // (i.e. make the black level 16*235 and the white level 235*256)
+    double yBlackLevel = blackIreLevel;
+    double yScale = 219.0 * 257.0 / (whiteIreLevel - blackIreLevel);
+
+    // Compute I & Q scaling factors
+    const double iqScale = chromaGain / (whiteIreLevel - blackIreLevel);
+    const double cbScale = 112*256 / (ONE_MINUS_Kb * kB); // Poynton, Eq 25.5 & 28.1
```
atsampson, line 45: Spaces around *.

tools/ld-chroma-decoder/ycbcr.cpp
```diff
@@ -0,0 +1,98 @@
...
+        // This doesn't affect the chroma scaling.
+        yScale *= 125.0 / 100.0;
+    }
+
+    for (const YIQ *yiq = begin; yiq < end; yiq++) {
+        double y = yiq->y;
+        double i = yiq->i;
+        double q = yiq->q;
+
+if (0) {
```
atsampson, line 60: Dead code - remove.

tools/ld-chroma-decoder/ycbcr.cpp
```diff
@@ -0,0 +1,98 @@
...
+        // Scale and offset to create CbCr
+        double Cb = U * cbScale + 128*256;
+        double Cr = V * crScale + 128*256;
+
+        // Clamp to valid range, per ITU-R BT.601-7 � 2.5.3
+        y  = qBound(1.0*256.0, y,  254.75*256.0);
+        Cb = qBound(1.0*256.0, Cb, 254.75*256.0);
+        Cr = qBound(1.0*256.0, Cr, 254.75*256.0);
+
+if (0)
```
atsampson, line 91: Dead code.

tools/ld-chroma-decoder/ycbcr.h
```diff
@@ -0,0 +1,66 @@
...
+#ifndef YCBCR_H
+#define YCBCR_H
+
+#include <QCoreApplication>
+#include <QDebug>
+
+#include "yiq.h"
+
+// ITU-R BT.601-7
+#define ONE_MINUS_Kb (1 - 0.114)
```
atsampson, line 36: Don't use #define for this in C++ - make them static constexpr constants, so they're computed at compile time. Then you can specify the intended type, and you can write "static constexpr double SIN33 = sin(whatever);" rather than needing to write out the value.

tools/ld-chroma-decoder/main.cpp
```diff
@@ -449,12 +474,16 @@ int main(int argc, char *argv[])
...
         decoder.reset(new NtscDecoder(combConfig));
     } else if (decoderName == "mono") {
-        decoder.reset(new MonoDecoder);
+        decoder.reset(new MonoDecoder(combConfig));
     } else {
         qCritical() << "Unknown decoder " << decoderName;
         return -1;
     }
 
+    decoder->config.pix_fmt = combConfig.outputYCbCr ?
```
atsampson, line 483: As above, this shouldn't be necessary. Have a pixelFormat field in Decoder::Configuration using an enumerated type, and set it above with the other configuration that gets passed to the Decoder constructors - directly modifying it here isn't very nice.

atsampson (changes requested):
Thanks for this - the overall approach looks good, and it's nice to see references for the maths!

I've added a few review comments - most of these are minor stylistic things, but you can also simplify how the Configuration stuff works (by making MonoDecoder work like the PAL/NTSC decoders).

atsampson:
Incidentally - have you tried my testsuite for ld-chroma-decoder with your changes? If you close [this repo](https://github.com/atsampson/ld-decode-testsuite), the try-chroma-decoder script has a fairly extensive set of tests, many of which can work using the files in ld-decode-testdata; it should produce the same output before your changes and after.

ifb:
@atsampson Do you want changes as a separate patch or squashed into this one?  If I `push -f` to my branch, I'm not sure what GitHub does to the merge request and your comments.  I could always squash later?

tools/ld-chroma-decoder/ycbcr.cpp
ifb: Whoops.  I knew I'd forget to remove that.  Be comforted that 20 IRE +Q and -I vectors do decode correctly. :)  +Q has negative green and -I has negative red when you decode to RGB, so it's a good test case.

tools/ld-chroma-decoder/ycbcr.cpp
ifb: Ditto.

tools/ld-analyse/tbcsource.cpp
```diff
@@ -536,7 +536,7 @@ QImage TbcSource::generateQImage(qint32 frameNumber)
         // Chroma decode the current frame and display
 
         // Decode colour for the current frame, to RGB 16-16-16 interlaced output
-        QVector<RGBFrame> outputFrames(1);
+        QVector<videoFrame> outputFrames(1);
```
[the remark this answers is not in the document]
ifb: I like OutputFrame.

tools/ld-chroma-decoder/comb.cpp
```diff
@@ -729,3 +729,40 @@ void Comb::FrameBuffer::overlayMap(const FrameBuffer &previousFrame, const Frame
...
+videoFrame Comb::FrameBuffer::yiqToYUVFrame()
+{
+    videoFrame outputFrame;
+
+    outputFrame.Y.resize(videoParameters.fieldWidth * frameHeight);
+    outputFrame.Cb.resize(videoParameters.fieldWidth * frameHeight);
+    outputFrame.Cr.resize(videoParameters.fieldWidth * frameHeight);
+
+    // Initialise the output frame
+    outputFrame.Y.fill(16*256);
```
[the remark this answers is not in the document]
ifb: I'm so used to `if( expr );`   I realized part way through I wasn't being consistent, but thought I had fixed most of it.

tools/ld-chroma-decoder/comb.h
```diff
@@ -35,7 +35,8 @@
 #include "lddecodemetadata.h"
 
 #include "rgb.h"
-#include "rgbframe.h"
+#include "ycbcr.h"
```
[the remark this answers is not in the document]
ifb: Yeah, I wasn't sure what the rationale for include ordering was, so I deferred to minimized changes.  I can sort them.

tools/ld-chroma-decoder/decoder.h
```diff
@@ -83,14 +83,16 @@ class Decoder {
...
+        bool outputYCbCr = false;
+    } config;
```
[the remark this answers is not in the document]
ifb: I'm a C person (can't you tell?), so most of this feels icky, but even this felt super icky to me.  We'll have to discuss on IRC/Discord.  I tried without it and couldn't get it to work.   Comb, Decoder, and Decoderpool all needed to know if we were doing YCbCr or not.

tools/ld-chroma-decoder/monodecoder.cpp
ifb: At one point I passed it a palConfig.  All of the option setting in main.cpp is through palConfig and combConfig, but combConfig isn't an Encoder subclass so I don't know how to unify it all nicely.

tools/ld-chroma-decoder/decoder.h
ifb: I copied pix_fmt libavformat and didn't think twice about a proper C++ name.

tools/ld-chroma-decoder/main.cpp
ifb: I thought I tested setting both, but maybe not or I didn't try reordering.  Anyway, the next patch will be to add y4m output (yuv with headers), so having a more general output-format option would be good.

I'd argue YCbCr should be the default, but I don't want to break everyone's scripts.

tools/ld-chroma-decoder/palcolour.cpp
ifb: So much whitespace.   All of the *256 mess is to be able to use 8-bit constants instead of magic 16-bit numbers that nobody remembers where they came from.

tools/ld-chroma-decoder/ycbcr.h
ifb: I figured there was a C++ way to do this at compile time without using the preprocessor.

tools/ld-chroma-decoder/main.cpp
ifb: Yes.  I felt mean doing it.

tools/ld-chroma-decoder/ycbcr.h
ifb: constexpr trig functions don't seem to be a thing, so I had to keep the magic values.

tools/ld-chroma-decoder/main.cpp
ifb: Dropped `--rgb` and `--yuv`

Use `--output-format` and `-p` (for pixel?).

ifb:
> Incidentally - have you tried my testsuite for ld-chroma-decoder with your changes? If you clone [this repo](https://github.com/atsampson/ld-decode-testsuite), the try-chroma-decoder script has a fairly extensive set of tests, many of which can work using the files in ld-decode-testdata; it should produce the same output before your changes and after.

I cloned it a long time ago, but never used it since I didn't have any of the test LD files.  I recently downloaded GGV1011 so I'd have something that's PAL, so `try-chroma-decoder` is running with that right now.  I don't know if my laptop can decode it before I have to leave with it.

atsampson (approved):
This all looks good to me now - thank you very much!

I've just run my testsuite over it and it produces bitwise identical output to master for all the existing tests (so I guess I'll have to write some YCbCr tests now...).

 I'd suggest squashing the revisions into your original changes in this case - it doesn't make much difference here but it makes the history a bit tidier if anyone needs to do a bisection in the future.

PR #619, opened 2021-02-18
Add yuv4mpeg (Y4M) output

ifb:
ld-chroma-decoder: add 'y4m' to the available --output-format options.

'y4m' output is identical to 'yuv' except with the addition of headers. This
avoids having to specify the dimensions, frame rate, field order, aspect ratio,
and pixel format to downstream consumers such as ffmpeg.

For example:
ld-chroma-decoder -p y4m input - | ffmpeg -f yuv4mpegpipe -i - output.mkv

### DIFF
--- a/tools/ld-chroma-decoder/comb.h
+++ b/tools/ld-chroma-decoder/comb.h
@@ -57,6 +57,7 @@ public:
         bool showMap = false;
         Decoder::PixelFormat pixelFormat = Decoder::PixelFormat::RGB48;
         bool outputYCbCr = false;
+        bool outputY4m = false;
 
         double cNRLevel = 0.0;
         double yNRLevel = 1.0;
--- a/tools/ld-chroma-decoder/decoder.h
+++ b/tools/ld-chroma-decoder/decoder.h
@@ -35,6 +35,14 @@
 #include "outputframe.h"
 #include "sourcefield.h"
 
+static const QString Y4M_CS_YUV444P16 = QStringLiteral(" C444p16 XCOLORRANGE=LIMITED\n");
+static const QString Y4M_CS_GRAY16    = QStringLiteral(" Cmono16 XCOLORRANGE=LIMITED\n");
+static const QString Y4M_PAR_NTSC_43  = QStringLiteral("97:114");  // (4 / 3) * (485 / 760)
+static const QString Y4M_PAR_NTSC_169 = QStringLiteral("194:171"); // (16 / 9) * (485 / 760)
+static const QString Y4M_PAR_PAL_43   = QStringLiteral("384:461"); // (4 / 3) * (576 / 922)
+static const QString Y4M_PAR_PAL_169  = QStringLiteral("512:461"); // (16 / 9) * (576 / 922)
+static constexpr QChar y4mFieldOrder = 't';
+
 class DecoderPool;
 
 // Abstract base class for chroma decoders.
@@ -83,8 +91,14 @@ public:
         GRAY16
     };
 
-    // After configuration, return a readable output pixel format
+    // Return a readable output pixel format
     virtual const char *getPixelName() const = 0;
+
+    // Return true if the decoder will output Y4M
+    virtual bool isOutputY4m() = 0;
+
+    // Generate the Y4M headers
+    virtual QString getHeaders() const = 0;
 
     // Parameters used by the decoder and its threads.
     // This may be subclassed by decoders to add extra parameters.
@@ -95,6 +109,7 @@ public:
         qint32 bottomPadLines;
         Decoder::PixelFormat pixelFormat = RGB48;
         bool outputYCbCr = false;
+        bool outputY4m = false;
     };
 
     // Compute the output frame size in Configuration, adjusting the active
--- a/tools/ld-chroma-decoder/decoderpool.cpp
+++ b/tools/ld-chroma-decoder/decoderpool.cpp
@@ -97,6 +97,13 @@ bool DecoderPool::process()
         }
     }
 
+    if (decoder.isOutputY4m()) {
+        if (targetVideo.write(decoder.getHeaders().toUtf8()) == -1) {
+            qCritical() << "Could not write Y4M header";
+            return false;
+        }
+    }
+
     qInfo() << "Using" << maxThreads << "threads";
     qInfo() << "Processing from start frame #" << startFrame << "with a length of" << length << "frames";
 
@@ -210,6 +217,12 @@ bool DecoderPool::putOutputFrame(qint32 frameNumber, const OutputFrame &outputFr
 
         // Save the frame data to the output file
         if (outputData.Y.size()) {
+            if (decoder.isOutputY4m()) {
+                if (targetVideo.write("FRAME\n") == -1) {
+                    qCritical() << "Writing to the output video file failed";
+                    return false;
+                }
+            }
             if (targetVideo.write(reinterpret_cast<const char *>(outputData.Y.data()), outputData.Y.size() * 2) == -1 ||
                 targetVideo.write(reinterpret_cast<const char *>(outputData.Cb.data()), outputData.Cb.size() * 2) == -1 ||
                 targetVideo.write(reinterpret_cast<const char *>(outputData.Cr.data()), outputData.Cr.size() * 2) == -1) {
--- a/tools/ld-chroma-decoder/monodecoder.cpp
+++ b/tools/ld-chroma-decoder/monodecoder.cpp
@@ -47,6 +47,31 @@ const char *MonoDecoder::getPixelName() const
     return config.outputYCbCr ? "GRAY16" : "RGB48";
 }
 
+bool MonoDecoder::isOutputY4m()
+{
+    return config.outputY4m;
+}
+
+QString MonoDecoder::getHeaders() const
+{
+    QString y4mHeader;
+    qint32 rateN = 30000;
+    qint32 rateD = 1001;
+    qint32 width = config.videoParameters.activeVideoEnd - config.videoParameters.activeVideoStart;
+    qint32 height = config.topPadLines + config.bottomPadLines +
+                    config.videoParameters.lastActiveFrameLine - config.videoParameters.firstActiveFrameLine;
+    QString y4mPixelAspect = (config.videoParameters.isWidescreen ? Y4M_PAR_NTSC_169 : Y4M_PAR_NTSC_43);
+    if (config.videoParameters.isSourcePal) {
+        rateN = 25;
+        rateD = 1;
+        y4mPixelAspect = (config.videoParameters.isWidescreen ? Y4M_PAR_PAL_169 : Y4M_PAR_PAL_43);
+    }
+    QTextStream(&y4mHeader) << "YUV4MPEG2 W" << width << " H" << height << " F" << rateN << ":" << rateD
+                            << " I" << y4mFieldOrder << " A" << y4mPixelAspect
+                            << (config.pixelFormat == YUV444P16 ? Y4M_CS_YUV444P16 : Y4M_CS_GRAY16);
+    return y4mHeader;
+}
+
 QThread *MonoDecoder::makeThread(QAtomicInt& abort, DecoderPool& decoderPool) {
     return new MonoThread(abort, decoderPool, config);
 }
--- a/tools/ld-chroma-decoder/comb.cpp
+++ b/tools/ld-chroma-decoder/comb.cpp
@@ -736,13 +736,17 @@ OutputFrame Comb::FrameBuffer::yiqToYUVFrame()
     OutputFrame outputFrame;
 
     outputFrame.Y.resize(videoParameters.fieldWidth * frameHeight);
-    outputFrame.Cb.resize(videoParameters.fieldWidth * frameHeight);
-    outputFrame.Cr.resize(videoParameters.fieldWidth * frameHeight);
-
-    // Initialise the output frame
     outputFrame.Y.fill(16 * 256);
-    outputFrame.Cb.fill(128 * 256);
-    outputFrame.Cr.fill(128 * 256);
+
+    if (configuration.chromaGain > 0) {
+        outputFrame.Cb.resize(videoParameters.fieldWidth * frameHeight);
+        outputFrame.Cr.resize(videoParameters.fieldWidth * frameHeight);
+        outputFrame.Cb.fill(128 * 256);
+        outputFrame.Cr.fill(128 * 256);
+    } else {
+        outputFrame.Cb.clear();
+        outputFrame.Cr.clear();
+    }
 
     // Initialise YIQ to YCbCr converter
     YCbCr ycbcr(videoParameters.white16bIre, videoParameters.black16bIre, configuration.whitePoint75, configuration.chromaGain);
--- a/tools/ld-chroma-decoder/main.cpp
+++ b/tools/ld-chroma-decoder/main.cpp
@@ -338,9 +338,11 @@ int main(int argc, char *argv[])
             monoConfig.outputY4m = true;
         }
         palConfig.outputYCbCr = true;
-        palConfig.pixelFormat = Decoder::PixelFormat::YUV444P16;
+        palConfig.pixelFormat = palConfig.chromaGain > 0 ? Decoder::PixelFormat::YUV444P16 :
+                                                           Decoder::PixelFormat::GRAY16;
         combConfig.outputYCbCr = true;
-        combConfig.pixelFormat = Decoder::PixelFormat::YUV444P16;
+        combConfig.pixelFormat = combConfig.chromaGain > 0 ? Decoder::PixelFormat::YUV444P16 :
+                                                             Decoder::PixelFormat::GRAY16;
         monoConfig.outputYCbCr = true;
         monoConfig.pixelFormat = Decoder::PixelFormat::GRAY16;
     } else if (outputFormatName != "rgb") {
--- a/tools/ld-chroma-decoder/main.cpp
+++ b/tools/ld-chroma-decoder/main.cpp
@@ -154,7 +154,7 @@ int main(int argc, char *argv[])
 
     // Option to select the output format (-p)
     QCommandLineOption outputFormatOption(QStringList() << "p" << "output-format",
-                                       QCoreApplication::translate("main", "Output format (rgb, yuv; default rgb); RGB48, YUV444P16, GRAY16 pixel formats are supported"),
+                                       QCoreApplication::translate("main", "Output format (rgb, yuv, y4m; default rgb); RGB48, YUV444P16, GRAY16 pixel formats are supported"),
                                        QCoreApplication::translate("main", "output-format"));
     parser.addOption(outputFormatOption);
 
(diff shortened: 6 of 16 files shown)


Gamnn:
Getting an `Operation not permitted` error when putting it into ffmpeg (4.2.4-1ubuntu0.1)
It also fails in mpv 0.32.0 which uses ffmpeg libraries.

Header is:
```
YUV4MPEG2 W760 H488 F30000:1001 It A1952:2280 C444p16 XYSCSS=444P16 XCOLORRANGE=LIMITED
FRAME
```

I compared it to vspipe's y4m header, which does work:
```
YUV4MPEG2 C444p16 W760 H488 F30000:1001 Ip A0:0 XLENGTH=53563
FRAME
```

After removing XCOLORRANGE=LIMITED from the header, both ffmpeg and mpv are happy.

```
ffmpeg -debug -f yuv4mpegpipe -i test.y4m test.mkv
ffmpeg version 4.2.4-1ubuntu0.1 Copyright (c) 2000-2020 the FFmpeg developers
  built with gcc 9 (Ubuntu 9.3.0-10ubuntu2)
  configuration: --prefix=/usr --extra-version=1ubuntu0.1 --toolchain=hardened --libdir=/usr/lib/x86_64-linux-gnu --incdir=/usr/include/x86_64-linux-gnu --arch=amd64 --enable-gpl --disable-stripping --enable-avresample --disable-filter=resample --enable-avisynth --enable-gnutls --enable-ladspa --enable-libaom --enable-libass --enable-libbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libcodec2 --enable-libflite --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libgme --enable-libgsm --enable-libjack --enable-libmp3lame --enable-libmysofa --enable-libopenjpeg --enable-libopenmpt --enable-libopus --enable-libpulse --enable-librsvg --enable-librubberband --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libtwolame --enable-libvidstab --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx265 --enable-libxml2 --enable-libxvid --enable-libzmq --enable-libzvbi --enable-lv2 --enable-omx --enable-openal --enable-opencl --enable-opengl --enable-sdl2 --enable-libdc1394 --enable-libdrm --enable-libiec61883 --enable-nvenc --enable-chromaprint --enable-frei0r --enable-libx264 --enable-shared
  libavutil      56. 31.100 / 56. 31.100
  libavcodec     58. 54.100 / 58. 54.100
  libavformat    58. 29.100 / 58. 29.100
  libavdevice    58.  8.100 / 58.  8.100
  libavfilter     7. 57.100 /  7. 57.100
  libavresample   4.  0.  0 /  4.  0.  0
  libswscale      5.  5.100 /  5.  5.100
  libswresample   3.  5.100 /  3.  5.100
  libpostproc    55.  5.100 / 55.  5.100
 matched as AVOption 'debug' with argument '-f'.
Reading option 'yuv4mpegpipe' ... matched as output url.
Reading option '-i' ... matched as input url with argument 'test.y4m'.
Reading option 'test.mkv' ... matched as output url.
Finished splitting the commandline.
Parsing a group of options: global .
Successfully parsed a group of options.
Parsing a group of options: input url test.y4m.
Successfully parsed a group of options.
Opening an input file: test.y4m.
[NULL @ 0x55948d669880] Opening 'test.y4m' for reading
[file @ 0x55948d66a280] Setting default whitelist 'file,crypto'
[yuv4mpegpipe @ 0x55948d669880] Format yuv4mpegpipe probed with size=2048 and score=100
[AVIOContext @ 0x55948d672640] Statistics: 32768 bytes read, 0 seeks
test.y4m: Operation not permitted

```

```
mpv -v test.y4m
[cplayer] Command line options: '-v' 'test.y4m'
[cplayer] mpv 0.32.0 Copyright © 2000-2020 mpv/MPlayer/mplayer2 projects
[cplayer]  built on UNKNOWN
[cplayer] ffmpeg library versions:
[cplayer]    libavutil       56.31.100
[cplayer]    libavcodec      58.54.100
[cplayer]    libavformat     58.29.100
[cplayer]    libswscale      5.5.100
[cplayer]    libavfilter     7.57.100
[cplayer]    libswresample   3.5.100
[cplayer] ffmpeg version: 4.2.4-1ubuntu0.1
[cplayer] 
[cplayer] Configuration: ./waf configure --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu --confdir=/etc/mpv --zshdir=/usr/share/zsh/vendor-completions --enable-cdda --enable-dvdnav --enable-libmpv-shared --enable-libsmbclient --enable-sdl2 --enable-sndio --disable-build-date --enable-dvbin
[cplayer] List of enabled features: 52arch aligned_alloc alsa asm atomics caca cdda cplayer cplugins debug-build drm drmprime dvbin dvdnav egl egl-drm egl-helpers egl-x11 fchmod ffmpeg gbm gbm.h gl gl-wayland gl-x11 glibc-thread-name glob glob-posix gnuc gpl iconv jack jpeg lcms2 libarchive libass libass-osd libav-any libavcodec libavdevice libavutil libbluray libdl libm libmpv-shared librt libsmbclient linux-fstatfs lua memfd_create optimize oss-audio plain-gl posix posix-or-mingw posix-spawn posix-spawn-native pthreads pulse rubberband sdl2 sdl2-audio sdl2-gamepad sdl2-video sndio stdatomic uchardet vaapi vaapi-drm vaapi-egl vaapi-wayland vaapi-x-egl vaapi-x11 vdpau vdpau-gl-x11 vt.h wayland wayland-protocols x11 xv zlib
[cplayer] Reading config file /etc/mpv/encoding-profiles.conf
[cplayer] Applying profile 'default'...
[cplayer] Reading config file /etc/mpv/mpv.conf
[cplayer] Applying profile 'default'...
[cplayer] Setting option 'hwdec' = 'vaapi' (flags = 4)
[cplayer] Reading config file /home/gamn/.config/mpv/mpv.conf
[cplayer] Applying profile 'default'...
[cplayer] Setting option 'hwdec' = 'vaapi-copy' (flags = 4)
[cplayer] Setting option 'screenshot-format' = 'png' (flags = 4)
[cplayer] Setting option 'v' = '' (flags = 8)
[cplayer] Waiting for scripts...
[cplayer] Set property: shared-script-properties -> 1
[cplayer] Set property: shared-script-properties -> 1
[cplayer] Set property: shared-script-properties -> 1
[osd/libass] Shaper: FriBidi 0.19.7 (SIMPLE) HarfBuzz-ng 2.6.4 (COMPLEX)
[osd/libass] Setting up fonts...
[osd/libass] Using font provider fontconfig
[osd/libass] Done.
[cplayer] Done loading scripts.
[cplayer] Running hook: ytdl_hook/on_load
[ytdl_hook] ytdl:// hook 
[ytdl_hook] not a ytdl:// url 
[osd/libass] Shaper: FriBidi 0.19.7 (SIMPLE) HarfBuzz-ng 2.6.4 (COMPLEX)
[ifo_dvdnav] Opening test.y4m
[osd/libass] Setting up fonts...
[bdmv/bluray] Opening test.y4m
[file] Opening test.y4m
[demux] Trying demuxers for level=normal.
[lavf] Found 'yuv4mpegpipe' at score=100 size=2048.
[lavf] avformat_open_input() failed
[demux] Trying demuxers for level=unsafe.
[osd/libass] Using font provider fontconfig
[osd/libass] Done.
[lavf] Found 'yuv4mpegpipe' at score=100 size=2048.
[lavf] avformat_open_input() failed
[cplayer] Opening failed or was aborted: test.y4m
[cplayer] Running hook: ytdl_hook/on_load_fail
[ytdl_hook] full hook 
[cplayer] finished playback, unrecognized file format (reason 4)
[cplayer] Failed to recognize file format.
[cplayer] 
[cplayer] 
[cplayer] Exiting... (Errors when loading file)
[cplayer] Set property: shared-script-properties -> 1

```

atsampson:
It might be [this change](https://git.ffmpeg.org/gitweb/ffmpeg.git/patch/0b1ff3265e9bdad3b4b6b97ced2f126cb3599568), which increased the max header size in yuv4mpegdec from 80 bytes to 96 in ffmpeg 4.3 - the header quoted above is > 80 bytes long...

ifb:
Yep.  The current patch works as-is with ffmpeg master.  ffmpeg 4.2.4 shipped in Ubuntu 20.04 doesn't like the long header.  Removing XCOLORRANGE does not affect ffmpeg master, so we can simply remove XCOLORRANGE to shorten the Y4M header. 
 Old ffmpeg users just won't get to simplify their command lines as much.

Note that removing XCOLORRANGE causes 4.2.4 to ignore the field order too, so you get progressive output on top of no explicit range signaling in the container.  Not the end of the world and it maintains the status quo with 4.2.4 ,but users may want to upgrade ffmpeg for things to work a little more intelligently.

Also the Ubuntu build of 4.2.4 doesn't seem to support monochrome (X264_CSP_I400) in its libx264 wrapper.  It's present in libx264.c behind a #define, so maybe there's some magic to enabling it.  You would think setting -pix_fmt gray would be enough (it is with static builds from [here](https://johnvansickle.com/ffmpeg/)).

Gamnn:
Ah-hah. Perhaps it would be better to leave it in, and remove the aspect ratio from the header instead. That also seems make it short enough to work with the stock ffmpeg, and is easy to specify manually.

ifb:
The libx264 in Ubuntu 20.04 is too old to support monochrome, so that explains that.

The "X" options are technically comments, so I'm inclined to drop something like XCOLORSPACE over a "core" field like aspect ratio.  However, I understand adding `-aspect 4:3` is a little easier, given how brain-damaged the packaged ffmpeg is. 🤷 

Just throw a new ffmpeg binary in `/opt/bin` and you can stop caring. 😄

tools/ld-chroma-decoder/decoderpool.cpp
```diff
@@ -210,6 +210,18 @@ bool DecoderPool::putOutputFrame(qint32 frameNumber, const OutputFrame &outputFr
 
         // Save the frame data to the output file
         if (outputData.Y.size()) {
+            if (decoder.isOutputY4m()) {
```
atsampson, line 213: Can you write the header immediately after the file is opened in DecoderPool::process, rather than checking whether it's the first frame here? That'd also still include the header if no frames get output.

tools/ld-chroma-decoder/monodecoder.cpp
```diff
@@ -47,6 +47,32 @@ const char *MonoDecoder::getPixelName() const
     return config.outputYCbCr ? "GRAY16" : "RGB48";
 }
 
+bool MonoDecoder::isOutputY4m()
+{
+    return config.outputY4m;
+}
+
+QString MonoDecoder::getHeaders() const
```
atsampson, line 55: Can getHeaders be in the Decoder base class, rather than being replicated three times? The code looks the same in all three versions...

tools/ld-chroma-decoder/comb.h
```diff
@@ -57,6 +57,8 @@ class Comb
         bool showMap = false;
         Decoder::PixelFormat pixelFormat = Decoder::PixelFormat::RGB48;
         bool outputYCbCr = false;
+        bool outputY4m = false;
+        bool isFirstFieldFirst = true;
```
atsampson, line 61: You can avoid replicating isFirstFieldFirst here - it'll be the same as videoParameters.isFirstFieldFirst, which is already in this structure.

Gamnn:
> The "X" options are technically comments, so I'm inclined to drop something like XCOLORSPACE over a "core" field like aspect ratio. However, I understand adding `-aspect 4:3` is a little easier, given how brain-damaged the packaged ffmpeg is.

Well, my thought being is there are non-4:3 aspect LDs as well, and it's not a vital parameter for y4m. Whichever you think is better for compatibility though.

tools/ld-chroma-decoder/monodecoder.cpp
ifb: I started out doing that, but the base class doesn't have a Configuration struct.  I gave it a private Decoder::Configuration config, but my C++ skillz are lacking and Decoder::getHeaders() returns nonsense, despite me setting config.xxxx in the derived classes.

atsampson: Oh, yes - the base class can't have a Configuration member because it needs to be one of the subclasses. Leave it as is for now then and I'll have a think...

ifb:
Good point about anamorphic content.  Setting aspect ratio to unknown seems reasonable.  Alternatively, could add an anamorphic flag to ld-chroma-decode?

Oh, and I didn't try to reduce the fraction.  That would save some header space, but maybe not enough to matter?

atsampson:
We could have a field in the .tbc.json for aspect ratio, and assume 4:3 otherwise - the VBI decoder could pick it up from WSS automatically. (It's also another thing that padding makes tricky, because the output isn't actually 4:3 until you crop off the padding...)

Gamnn:
Yeah. I tried `YUV4MPEG2 W760 H488 F30000:1001 It A4:3 C444p16 XYSCSS=444P16 XCOLORRANGE=LIMITED` but it seems to be just 1 character too long. 

It's a shame they didn't choose a shorter parameter name than colorrange given the header size constraints.

But since we're supporting Ubuntu 20.04 LTS, it's probably best to keep it under 80 for the time being. So gotta cut one or the other. Though, what does XYSCSS do? Why is it needed in addition to the C444p16?

ifb:
Separate from this patch, I'd argue for mod2 instead of mod8 padding, in part because it borks the aspect ratio.

But that's a separate discussion for later.

ifb:
> Yeah. I tried `YUV4MPEG2 W760 H488 F30000:1001 It A4:3 C444p16 XYSCSS=444P16 XCOLORRANGE=LIMITED` but it seems to be just 1 character too long.
> 
> It's a shame they didn't choose a shorter parameter name than colorrange given the header size constraints.
> 
> But since we're supporting Ubuntu 20.04 LTS, it's probably best to keep it under 80 for the time being. So gotta cut one or the other. Though, what does XYSCSS do? Why is it needed in addition to the C444p16?

It's a limitation of ffmpeg's parser that has been fixed.  There's nothing in the format that limits the header size.  It's an implementation detail in 2.4.3.

Also 'A4:3' isn't correct.  That field is for the SAR, not DAR.

XYSCSS is because the original mjpegtools implementation of Y4M only had a few formats supported.  Everything else is a semi-standard extension.

https://git.ffmpeg.org/gitweb/ffmpeg.git/blob_plain/HEAD:/libavformat/yuv4mpegenc.c
https://wiki.multimedia.cx/index.php/YUV4MPEG2

tools/ld-chroma-decoder/comb.h
ifb: isFirstFieldFirst is a private member of LdDecodeMetaData.

I tried using getIsFirstFieldFirst() originally, but I only have the LdDecodeMetaData::videoParameters struct available.

Gamnn:
> It's a limitation of ffmpeg's parser that has been fixed. There's nothing in the format that limits the header size. It's an implementation detail in 2.4.3.

Well, even if not inherent in the format, ffmpeg's implementation is still limited to 96 bytes in master. Otherwise it might have been nice to also include the XLENGTH parameter like vspipe does.

tools/ld-chroma-decoder/comb.h
ifb: Per our discussion on IRC, this ends up being constant since any potential field swapping will be done on the input to ld-chroma-decode.

tools/ld-chroma-decoder/decoderpool.cpp
ifb: Done.

ifb:
Updated pull request.  Added a field to the .json for signaling widescreen, which is always false for now.

The Y4M pixel aspect ratios are calculated before mod8 padding  This makes the DAR correct no matter what padding there may be, even if it isn't exactly 4:3.

ifb:
I spoke too soon.  With ffmpeg 4.2.4, adding XLENGTH works for PAL content if it is under 10,000 frames but NTSC is broken no matter what.

@Gamnn What actually uses XLENGTH?  vspipe writes it, but if x264 and ffmpeg don't use it, what does?

Gamnn:
> I spoke too soon. With ffmpeg 4.2.4, adding XLENGTH works for PAL content if it is under 10,000 frames but NTSC is broken no matter what.

I was worried CLV discs might push it over the limit.
 
> @Gamnn What actually uses XLENGTH? vspipe writes it, but if x264 and ffmpeg don't use it, what does?

🤷 Nothing I currently use currently does, so it was just a "might be nice." I'd care more about it if ffmpeg did use it- always nice to see an ETA.

oyvindln:
If there are some variations of options that don't work with older ffmpeg, one option is to have an extra option that adds the additional settings that are convenient but won't work with all version.

ifb:
> If there are some variations of options that don't work with older ffmpeg, one option is to have an extra option that adds the additional settings that are convenient but won't work with all version.

We can fit everything that matters.  XYSCSS is for legacy compatibility before the additions to XCOLORSPACE were "standardized."  XLENGTH is written by vspipe, but doesn't seem to be read by anything.  If we drop both, we can easily fit within the stock ffmpeg's 80-byte header limit.

Using ffmpeg master gets you a more automatic settings for field order and limited/full range, but even it's not consistent with different codecs (ffv1 will automatically encode interlaced, libx264 will not).  Oh, well.  Dimensions, frame rate, and colorspace are the big ones and those work fine now.

Gamnn:
I noticed that while -f mono does set it to Cmono16, using --blackandwhite with another decoder doesn't. Other than that, it's looking good.

ifb:
> I noticed that while -f mono does set it to Cmono16, using --blackandwhite with another decoder doesn't. Other than that, it's looking good.

That's expected behavior, the same as  `--output-format yuv`.  I considered doing as you suggest, but aside from the slightly added complexity, I thought it might be a little confusing if the decoder pixel format changed based on other flags.  For now, only mono outputs grayscale, but I don't have a strong opinion about it.  Of course, confusion is less of an issue if you're using y4m and all of this gets signaled automatically.

`--blackandwhite` is an alias for `--chroma-gain 0`.  Other than debugging, is either more useful than `-f mono`?

Gamnn:
> `--blackandwhite` is an alias for `--chroma-gain 0`. Other than debugging, is either more useful than `-f mono`?
> 
> Couldn't `--blackandwhite` just as easily be an alias for `-f mono`?

`-f mono` goes through no y/c separation, so it will leave chroma dots in the output if there's any color information in the source. But it's ideal for sources that have no color information as it will be free of any defects caused by chroma decoders.
Using `--blackandwhite` or `--chroma-gain 0` with another decoder goes through the separation, then discards the chroma, leaving the output free of chroma dots. A use case for this might be a black & white film with a color cast.
It's not a huge deal, but since it's only luma at this point it would make sense to also output it at gray16.

ifb:
Patchset updated.

`--blackandwhite` outputs GRAY16 now.

tools/ld-chroma-decoder/decoder.h
```diff
@@ -35,6 +35,14 @@
 #include "outputframe.h"
 #include "sourcefield.h"
 
+static QString Y4M_CS_YUV444P16 = QStringLiteral(" C444p16 XCOLORRANGE=LIMITED\n");
+static QString Y4M_CS_GRAY16    = QStringLiteral(" Cmono16 XCOLORRANGE=LIMITED\n");
+static QString Y4M_PAR_NTSC_43  = "97:114";  // (4 / 3) * (485 / 760)
+static QString Y4M_PAR_NTSC_169 = "194:171"; // (16 / 9) * (485 / 760)
+static QString Y4M_PAR_PAL_43   = "384:461"; // (4 / 3) * (576 / 922)
+static QString Y4M_PAR_PAL_169  = "512:461"; // (16 / 9) * (576 / 922)
+static constexpr QChar y4mFieldOrder = 't';
```
oyvindln, line 44: May want to make these static const as they're not intended to be modified.

ifb:
Rebased.  Anything left I need to address?

atsampson (approved):
Looks good to me.

PR #629, opened 2021-04-13
Build cleanly with Qt 5.15 (and other minor fixes)

atsampson:
See the commit messages for more details. The general aim here was to get rid of the warnings when building ld-decode against Qt 5.15, but I found a few other trivial things in the process.

One warning remains, about `QLabel::pixmap()` being deprecated, but I can't fix that without adding a version conditional for older Qt and I don't think it's worth it.


atsampson:
Reviewed by committee on IRC :-)

PR #634, opened 2021-04-19
Upstreaming tweaks from vhs-decode

oyvindln:
Chroma decoder and convenience tweaks from vhs-decode fork - thought I'd upstream to keep somewhat in sync.

- Try loading a <file>.tbc.json if loading a <file>_chroma.tbc file and <file>_chroma.tbc.json doesn't exist
- Add ntsc-phase-comp decoder and options for it in ld-analyse and auto-enable on loading _chroma.tbc
  - Not sure why but had to rotate phase 10 degrees for it to look correct compared to normal decoder.
  - Chroma ends up 1 px off position relative to normal one but not sure which is actually correct, otherwise it's simple enough to compensate for
- chroma_lpf option in ld-chroma-decoder to enable the option, already present in ld-analyse (ended up not needing it but left it in)

### DIFF
--- a/tools/ld-analyse/chromadecoderconfigdialog.cpp
+++ b/tools/ld-analyse/chromadecoderconfigdialog.cpp
@@ -148,6 +148,8 @@ void ChromaDecoderConfigDialog::updateDialog()
 
     const bool isSourceNtsc = !isSourcePal;
 
+    ui->phaseCompCheckBox->setEnabled(isSourceNtsc);
+    ui->phaseCompCheckBox->setChecked(ntscConfiguration.phaseCompensation);
     ui->ntscFilter2DRadioButton->setEnabled(isSourceNtsc);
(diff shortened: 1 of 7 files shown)


atsampson:
Cheers for this. Please can we hold off merging it until after the work I'm doing in [this branch](https://github.com/atsampson/ld-decode/tree/outputframe) goes through? I'm reworking how the decoders produce output, which will make it easier to implement the rotation here - I can rebase this on top of that work once it's done.

I wouldn't worry about adding a command-line option for the regular IQ filter - it doesn't produce useful output at the moment, and I'm not sure it's really very useful anyway. (The demodulator in Comb doesn't need a post-filter, so it's only if you want to force the early NTSC filter specs.)

oyvindln:
Yeah that''s fine, need to tweak things a little anyhow.

oyvindln:
Closing for now, need to rework this.

Issue #570, opened 2020-11-23
Average/median waveform or vectoroscope display in the DD-Analyse GUI


Zcooger:
This ticket is intended to suggest adding checkbox that enables waveform or vectoroscope display averaging for a certain amount of frames from 2 to total of TBC file.
This would allow to cancel out noise and precisely set IRE or burst level .json values while looking at VITS generator lines.

simoninns:
It's hard to understand the intent here.  If I understand correctly, you are asking for the ability to show some form of graph which is an average of something over a number of frames...  Can you be a bit more specific?  What is it (over a range of 2 or more frames) that you want to see?  If it's only for a vectorscope add the details to issue #305 and indicate that this issue can be closed as duplicate.

Zcooger:
Example of such function: https://youtu.be/KWeY2yqwVA0 - 7:28

simoninns:
The clip is an FFT function...  It still doesn't clarify what you are asking for.  You want some type of graph-based output of some type of data to animate over a specified number of frames?  Or is this some type of static average over a specified number of frames of some type of data?

Zcooger:
That "average" button enables temporal noise cancellation over time. In ld-analyse it can average current line in waveform display once decoded frames are changed back and forth. Once line is switched or "average" checkbox deselected current display might be resetted and line is displayed without filtering.

Issue #305, opened 2019-08-02
Show real chroma decoder output in ld-analyse line scope


atsampson:
At present, the ld-analyse line scope has its own simple filters that are used to show luma and chroma components on the scope. It could show the chroma decoder's output here instead, which would be useful for debugging the decoders.

This would be easier if the decoders provided YUV or YIQ output, so it would make sense to do this as part of #283.

atsampson:
YUV output from the chroma decoders would also mean we could implement a vectorscope, which would be very useful for adjusting chroma gain and phase.

atsampson:
@Gamnn suggested that we could also have a luma histogram for adjusting black/white levels along with this.

PR #635, opened 2021-04-21
Split decoding from output conversion

atsampson:
Previously, ld-chroma-decoder's PAL, NTSC and monochrome decoders all produced R'G'B' or (after @ifb's work in #618) Y'CbCr output directly. This made for a fair amount of duplication between the different decoders.

This set of changes adds a ComponentFrame class, which holds a complete frame of Y'UV data, and an OutputWriter class that can convert a ComponentFrame to any of the supported output formats (outputting just the selected active area, so there's no need to crop afterwards). The Decoder classes now just produce ComponentFrames, with the NTSC decoders rotating IQ to UV, and output is handled centrally.

Besides simplifying the decoders, this also means that unscaled Y'UV is now available to ld-analyse - for example, to show real Y/C on the line graph or to implement a vectorscope - see #305 and #570.

I've left chroma gain as part of the decoder configuration rather than bringing it into OutputWriter because (a) it can be done efficiently as part of the UV rotation and (b) ~~the next plan is to add~~ the last commit adds a chroma phase control, which needs to be done in the decoder (because it has to be before the V-switch for PAL).

I've checked that this produces (effectively) the same output as the old code, and that all the performance-critical bits vectorise properly with GCC. Testing with valgrind revealed some uninitialised reads in the noise reduction code so there's a fix for that included as well.

Edit: I should note that @ifb has already had a look through this and sent me some suggestions which I've incorporated.


atsampson:
Chroma phase control added.

oyvindln:
Just wanna note that for non-laserdisc things, especially CVBS, it would be nice if the code accommodates for that there can be sources like PAL-M, PAL-N and more obscure formats that are not 625-line PAL or 525-line NTSC. Would probably still be so that the sample rate 4fsc of whatever sub carrier frequency was used, so it's mainly about accounting for line count and width that can be different.

atsampson:
Yes - if there's enough interest in supporting other standards, we could replace isOutputPal in the metadata with a field specifying a standard name, and then the metadata reader can fill in the other details based on that.

(I wonder what a sensible sampling rate for mono-only standards like UK 405 and French 819 lines would be... it would be easy enough to generate composite samples of these with hacktv.)

PR #639, opened 2021-05-02
Update ld-ldf-reader for ffmpeg 4.5

atsampson:
This was based on an example that used the old `avcodec_decode_audio4` API, which will be removed in ffmpeg 4.5. The changes here just update it to match the latest version of the example, which uses `avcodec_receive_frame` instead, with the seeking changes re-added.

The `av_register_all` function will also be removed in 4.5, but it's still necessary to call it in older versions of ffmpeg to get the codecs registered, so I've added a version conditional around it.

I've tested this with ffmpeg 3.4 (as in Ubuntu 18.04) and 4.5 in development (current master).

### DIFF
--- a/ld-ldf-reader.c
+++ b/ld-ldf-reader.c
@@ -1,6 +1,10 @@
 /*
- * adapted/gutted from demuxing_decoding.c - copyright below:
- * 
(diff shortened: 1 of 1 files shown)


happycube (approved):

happycube:
Thanks!

PR #645, opened 2021-07-05
Fix deemp_low and deemp_high CLI parameters

Mikewando:


Mikewando:
`ld-decode` stores these parameters in `"deemp_coeff"` https://github.com/happycube/ld-decode/blob/212bb3151a9bd29fad9f64a93faf7e5a84145ce5/ld-decode#L305
So of course this mismatch could also be fixed by changing how `ld-decode` stores them to match how `core.py` expects them, but that seemed like it might still require `core.py` changes since I think you could end up with `0` as the default value for both, which is clearly not intended.

happycube (approved):

happycube:
Thanks!

PR #691, opened 2022-01-20
Add ld-cut support to the github test suite

happycube:
This adds ld-cut support using existing test data for both NTSC and PAL with the seek option used for both.


atsampson (approved):
Looks good to me!

PR #693, opened 2022-01-26
Avoid get_git_info failing if commands fail and add two checks to avoid mean of empty slice warning

oyvindln:
get_git_version can fail since branch/commit are not initialized in all branches.

the version file also doesn't have a ":" in it, so not sure what the intention was there.


happycube (approved):
Didn't think of the fallback (oops), and that empty slice warning's annoying too.  Thanks!

PR #706, opened 2022-02-19
ld-chroma-decoder: First/last field/frame line + padding options

MooglyGuy:
- Added options --ffll / --first_active_field_line, --lfll / --last_active_field_line, --ffrl / --first_active_frame_line, --lfrl / --last_active_frame_line.
- Added value and range validation for both PAL and NTSC modes.
- Added option --pad / --output-padding to optionally pad along X and Y from 1 (disabled) to 32 pixels.
- Made LdDecodeMetaData cache the current VideoParameters struct rather than always pulling it from JSON on every getVideoParameters call.

### DIFF
--- a/tools/ld-chroma-decoder/main.cpp
+++ b/tools/ld-chroma-decoder/main.cpp
@@ -277,7 +277,7 @@ int main(int argc, char *argv[])
 
     // Positional argument to specify output video file
     parser.addPositionalArgument("output", QCoreApplication::translate("main", "Specify output file (omit or - for piped output)"));
-    
+
     // Process the command line options and arguments given by the user
     parser.process(a);
 
@@ -347,7 +347,7 @@ int main(int argc, char *argv[])
             return -1;
         }
     }
-    
+
     if (parser.isSet(chromaGainOption)) {
         const double value = parser.value(chromaGainOption).toDouble();
         palConfig.chromaGain = value;
@@ -488,7 +488,7 @@ int main(int argc, char *argv[])
         qCritical() << "Can only show FFTs with the transform2d/transform3d decoders";
         return -1;
     }
-    
+
     // Select the decoder
     QScopedPointer<Decoder> decoder;
     if (decoderName == "pal2d") {
--- a/tools/ld-chroma-decoder/main.cpp
+++ b/tools/ld-chroma-decoder/main.cpp
@@ -174,6 +174,12 @@ int main(int argc, char *argv[])
                                        QCoreApplication::translate("main", "Output in black and white"));
     parser.addOption(setBwModeOption);
 
+    // Option to select output padding (-pad)
+    QCommandLineOption outputPaddingOption(QStringList() << "pad" << "output-padding",
+                                       QCoreApplication::translate("main", "Pad the output frame to a multiple of this many pixels on both axes (1 means no padding, maximum is 32)"),
+                                       QCoreApplication::translate("main", "number"));
+    parser.addOption(outputPaddingOption);
+
     // Option to select which decoder to use (-f)
     QCommandLineOption decoderOption(QStringList() << "f" << "decoder",
                                      QCoreApplication::translate("main", "Decoder to use (pal2d, transform2d, transform3d, ntsc1d, ntsc2d, ntsc3d, ntsc3dnoadapt, mono; default automatic)"),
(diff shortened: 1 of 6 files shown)


tools/library/tbc/lddecodemetadata.cpp
```diff
@@ -24,6 +24,25 @@
 
 #include "lddecodemetadata.h"
 
+// Default line parameters for PAL decoding
```
atsampson, line 27: Please could you keep the comments that were in getVideoParameters, explaining why these defaults were chosen?

tools/ld-chroma-decoder/main.cpp
```diff
@@ -247,7 +277,7 @@ int main(int argc, char *argv[])
 
     // Positional argument to specify output video file
     parser.addPositionalArgument("output", QCoreApplication::translate("main", "Specify output file (omit or - for piped output)"));
-
+
```
atsampson, line 280: Remove this whitespace?

tools/ld-chroma-decoder/main.cpp
```diff
@@ -317,7 +347,7 @@ int main(int argc, char *argv[])
             return -1;
         }
     }
-
+
```
atsampson, line 350: Whitespace again.

tools/ld-chroma-decoder/main.cpp
```diff
@@ -440,7 +488,7 @@ int main(int argc, char *argv[])
         qCritical() << "Can only show FFTs with the transform2d/transform3d decoders";
         return -1;
     }
-
+
```
atsampson, line 491: Whitespace...

tools/library/tbc/lddecodemetadata.cpp
```diff
@@ -485,6 +579,7 @@ void LdDecodeMetaData::clearFieldDropOuts(qint32 sequentialFieldNumber)
     json.remove({"fields", fieldNumber, "dropOuts", "fieldLine"});
 }
 
+
```
atsampson, line 582: Whitespace.

atsampson (changes requested):
Looks good to me - one comment to fix and a few whitespace errors introduced.

tools/library/tbc/lddecodemetadata.cpp
MooglyGuy: Addressed; comments re-added.

tools/ld-chroma-decoder/main.cpp
MooglyGuy: Fixed, along with at lines 350 and 491 in this file, and line 582 in lddecodemetadata.cpp.

Please note that the whitespacing in the source file is already inconsistent, as re-saving the relevant source files in an editor with "Remove trailing whitespace" enabled results in a considerable amount of diffs.

MooglyGuy:
All specified issues have been addressed.

atsampson (approved):

tools/ld-chroma-decoder/main.cpp
atsampson: I can only see one line with trailing whitespace in current master (in palencoder.cpp, so it's my fault).

tools/library/tbc/lddecodemetadata.cpp
```diff
@@ -28,15 +28,19 @@
...
 const qint32 LdDecodeMetaData::LineParameters::sDefaultPALLastActiveFrameLine = 620;
 const qint32 LdDecodeMetaData::LineParameters::sDefaultPALFieldHeightCheck = 313;
 
 // Default line parameters for NTSC decoding
 const qint32 LdDecodeMetaData::LineParameters::sMinNTSCFirstActiveFrameLine = 1;
 const qint32 LdDecodeMetaData::LineParameters::sDefaultNTSCFirstActiveFieldLine = 20;
 const qint32 LdDecodeMetaData::LineParameters::sDefaultNTSCLastActiveFieldLine = 259;
+// Interlaced line 40 is NTSC line 21 (the closed-caption line before the first active half-line)
 const qint32 LdDecodeMetaData::LineParameters::sDefaultNTSCFirstActiveFrameLine = 40;
+// Interlaced line 524 is NTSC line 263 (the last active half-line).
```
ifb, line 43: I suppose "interlaced" actually means "frame" or "interleaved" in this context?  I've also seen this called psuedo-progressive numbering.  Not going to bikeshed it, but it's weird terminology IMHO.

Anyway, I haven't looked at PAL hard, but I don't think the NTSC comments are correct.  Unless numbering starts from 0?

Line numbering spreadsheet:
https://docs.google.com/spreadsheets/d/1pzVHFusLCI7kys2GzK9BTk3w7G8zcLxgHs3DMsurF7g/edit#gid=117149856

ifb:
I have a local patch that's very similar, but I like this one better.  (I really should make some merge requests...)

Note that whenever you pad an odd number of lines at the top, it flips the field order which affects the y4m header.

SMPTE 170M line 263 is half active on the left and below line 525.  Line 283 is half active on the right and above line 21.  This is why 486-line NTSC is bottom field first (it starts at F2 L283 and ends at F1 L263).  I'd argue 486-line NTSC should be the default, but that's for another patch.

atsampson:
Line numbering in ld-decode is generally a bit of a mess - various places in the code use {0-based | 1-based} {within a field | within an interleaved frame }, and the confusion between the different conventions has resulted in several bugs in the past. In this case it's 0-based within the interleaved frame. About the only approach that isn't used anywhere is the NTSC/PAL standard numbering...

I've occasionally wondered about adding a LineNo class that can convert between the different representations, so we can at least show the standard numbering in the GUI.

tools/library/tbc/lddecodemetadata.cpp
MooglyGuy: I appreciate the feedback, but the change you're commenting on was simply me re-adding a comment that already existed prior to even my changes. So it should probably be given as a separate issue on Github, rather than on my pull request.

ifb: I didn't meant to imply that action from you was required.  It's more a question for happycube or ats about how the numbering works (which was answered below).

PR #714, opened 2022-04-04
Remove ellipsis from "Reload TBC" option

marnanel:
Ellipses mean there's a dialogue box, and in this case there isn't.


atsampson:
Looks good to me - thanks!

PR #718, opened 2022-04-28
add -qt5 option to Makefile

marnanel:
add -qt5 option to qmake line in Makefile
    
qt5-default no longer exists under Debian bullseye and Ubuntu 21.04

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ DESTDIR =
 
 # Tools you might want to override
 PYTHON3 ?= python3
-QMAKE ?= qmake
+QMAKE ?= qmake -qt5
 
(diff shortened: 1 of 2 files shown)


atsampson:
Hm - what happens if you build with the default `QMAKE=qmake`? The code should work fine with either Qt 5 or Qt 6 - I normally build with the latest release of Qt 6 these days...

simoninns (changes requested):
I agree with @atsampson - I don't think this is the right change as Qt6 should work too.  Please confirm this change is really required and/or change it as suggested - then I will accept the PR

simoninns:
Closing as inactive

PR #720, opened 2022-05-05
ld-compress tweaks

Gamnn:
- Remove OpenCL setup function which was too environment-specific and required super-user access. Refers to the ld-decode wiki instead.
- Workaround for verify function to work independent of openssl version.
- Prevent overwriting of existing .ldf files when using flaccl (ffmpeg prevents this internally)
- Smarter setting of compression level between ffmpeg-based and flaccl-based compression.
- Adds .flac.ldf to flaccl output, to differentiate from normal ogg container .ldf files.
- Fixes issue where it would fail to verify .ldf files in a flac container instead of ogg.

### DIFF
--- a/scripts/ld-compress
+++ b/scripts/ld-compress
@@ -5,33 +5,43 @@ comp=true
 uncomp=false
 verify=false
 gpu=false
-setup=false
 fileinput_method=cat
 level=11
 ext=ldf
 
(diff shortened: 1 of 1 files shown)


TokugawaHeavyIndustries:
Looks great

PR #721, opened 2022-05-14
Replace JsonWax with an event-based parser

### LINKED ISSUE #684: Replace JsonWax with RapidJSON
Rebuild the JSON metadata library functions using the RapidJSON library and use a SAX style interface to streamline the whole implementation.

atsampson:
Rework LdDecodeMetaData so that the JSON input is parsed directly into the structures that the tools use, rather than being parsed into JsonWax's QVariant-based tree and then converted repeatedly.

This is about six times faster for the new metadata test program with a typical CLV side. Things that previously ended up invoking the parser multiple times (e.g. dropout processing) may benefit more.

This would benefit from some review, since it makes a lot of changes to how the metadata code works - it produces identical output from my test suite, and I've tried it on a range of .tbc.json files produced over the last few years, but I wouldn't be surprised if there are still odd cases it doesn't get right.

Fixes #684.


simoninns (approved):
Looks like some amazing work :)  It requires some hard testing I guess - but probably better to roll it out and let 'everyone' test - it's such a core part of the library that any issues should pop up pretty fast.

happycube (approved):
Haven't looked at this much but sounds good to me - thanks!

PR #727, opened 2022-05-21
Use double values for sampleRates and fSC

### LINKED ISSUE #726: Remove videoParameters.fsc from JSON
Currently the JSON `videoParameters` object contains a field called `fsc`: "The fSC in Hz, truncated to an integer".

The "truncated" part is important, because for both NTSC and PAL the true value has a fractional part. The few places in the tools where it's used expect it to be the correct value, so they're very slightly wrong at the moment - and for PAL this does actually make a small difference to the output.

Since the value of fSC is fixed for each video standard we support, I'm going to suggest that we:
- stop ld-decode from writing `fsc` at all, and remove it from the documentation
- make `VideoParameters::fsc` a double, rather than an integer, and rename it to `fSC`
- make the JSON parser ignore the fsc field if present, and set `fSC` exactly based on the video standard (as it does for `LineParameters` and other per-standard information)

I'd also like to make the audio and video `sampleRate` parameters doubles at the same time - in all but one place in the code they're used in a double context, and there's no reason why you couldn't have a non-integer sample rate.

The tools will still handle existing JSON files correctly after this change.

Any thoughts?

atsampson:
Fixes #726. This is built on a slightly-refactored version of MooglyGuy's LineParameters work - the big changes are in the last commit:

Use double values for `sampleRate` and `fSC`. Several of the tools want exact values for these.

For video `sampleRate`, previously ld-decode output `int(fSC) * 4` to the JSON, while actually using `fSC * 4` internally. Output the full floating-point value of `fSC * 4` instead. Correct ld-chroma-encoder to match this.

The audio `sampleRate` has always been an integer anyway in ld-decode.

For `fsc`, previously ld-decode output `int(fSC)`. Since fSC is fixed for each video standard, there is no need to store it in the JSON; set it to the exact value in `initialiseVideoSystemParameters` along with the other system-dependent parameters.

This is backwards-compatible, in that the tools can still work with older JSON files.

(Why not have ld-decode continue to output `fsc` anyway, for backwards compatibility with older versions of the tools? Because comb.cpp checked that the integer sample rate was exactly 4 * the integer fSC -- and that isn't true for the correct NTSC values when rounded to integers.)

### DIFF
--- a/lddecode/core.py
+++ b/lddecode/core.py
@@ -3966,10 +3966,8 @@ class LDdecode:
 
         vp["isSourcePal"] = True if f.rf.system == "PAL" else False
 
-        vp["fsc"] = int(f.rf.SysParams["fsc_mhz"] * 1000000)
         vp["fieldWidth"] = f.rf.SysParams["outlinelen"]
-        vp["sampleRate"] = vp["fsc"] * 4
-        spu = vp["sampleRate"] / 1000000
+        vp["sampleRate"] = f.rf.SysParams["outfreq"] * 1000000
 
         vp["black16bIre"] = np.float(f.hz_to_output(f.rf.iretohz(self.blackIRE)))
         vp["white16bIre"] = np.float(f.hz_to_output(f.rf.iretohz(100)))
@@ -3978,6 +3976,7 @@ class LDdecode:
 
         # current burst adjustment as of 2/27/19, update when #158 is fixed!
         badj = -1.4
+        spu = f.rf.SysParams["outfreq"]
         vp["colourBurstStart"] = np.round(
             (f.rf.SysParams["colorBurstUS"][0] * spu) + badj
         )
--- a/tools/ld-analyse/tbcsource.cpp
+++ b/tools/ld-analyse/tbcsource.cpp
@@ -828,10 +828,6 @@ void TbcSource::startBackgroundLoad(QString sourceFilename)
     // Get the video parameters from the metadata
     LdDecodeMetaData::VideoParameters videoParameters = ldDecodeMetaData.getVideoParameters();
 
-    // Use default line parameters, as the user will not override it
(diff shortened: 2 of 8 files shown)


atsampson:
I'll update the JSON documentation when this is merged.

simoninns (approved):
Looks sane to me; I'll leave it to @happycube to review the ld-decode python changes.

oyvindln:
PAL/NTSC isn't strictly tied to line count. The variants used on laserdiscs are, but for composite and tapes, there is PAL-M, PAL-N, PAL60, NTSC 4.43 (latter 2 only really for composite) that were somewhat widely used, so ideally the decoder implementation ought to account for that.

The current  ld-chroma-decoder implementation already handles PAL-M from PAL-M tapes, so would be nice if we can avoid breaking that, I'll have to re-check if we added a separate json parameter, or if the code simply figures it based on the combination of pal and line count or other parameters in the JSON.

atsampson:
@oyvindln - Hm, yes, that's a good point. While in the long term I'd like the JSON metadata to specify the standard explicitly - i.e. replace `isSourcePal` with something like `"system": "PAL-M"`, so ld-chroma-decoder can use the right filters for each standard rather than assuming everything is standard PAL, I don't want to break how vhs-decode is currently representing PAL-M. I guess we do need to keep `fSC` in the JSON...

oyvindln:
It shouldn't need to be kept there, as long as ld-chroma-decoder has some way of determining 525 vs 625 line system besides isSourcePal in the json. Fieldheight, fieldwidth and sampleRate is still specified with this change, so the PAL decoder could set fsc accordingly depending on fieldheight to tell if it's standard PAL or PAL-M. (Can worry about PAL60 later, that's not really supported as of now anyway I think and only relevant for composite input.)

happycube (approved):

PR #728, opened 2022-05-21
Support multiple video systems, and add PAL-M support

atsampson:
CC @oyvindln, who kindly supplied a sample PAL-M VHS capture to test this with...

In the JSON, `isSourcePal` is replaced by a `"system"` string. In the library, this is converted to a `VideoSystem` enum; if `"system"` is not set, then the system is guessed from `isSourcePal` and `fieldHeight`, so the tools will still work with old files.

PAL-M uses PAL colour on a 525-line line standard with fSC being close to but not the same as NTSC -- so in some senses it works like standard PAL, and in some senses like standard NTSC. Adjust all the system-dependent parts of the tools as needed, including filters.

To make testing this easier, it also adds better support for vhs-decode's separate `_chroma.tbc` files in ld-analyse - it now loads both files if present, and lets you switch between Y, C or Y+C:

palmanalyse

(Note to self to update the JSON docs again if people are happy with this.)


happycube (approved):
approved for my tiny bit of core.py that got changed

simoninns (approved):
Looks good.  Seems like the ld-analyse user-guide is drifting way out of line with the actual code though.  Perhaps an issue should be raised for an update?

PR #717, opened 2022-04-28
Add commandline options for padding, and data output.

marnanel:
Add commandline option --pad. Add param on commandline to specify data output.
    
These work in non-interactive mode, too.

PR #729, opened 2022-05-27
Fix a comparison in ld-process-efm that doesn't work with Qt 6

atsampson:
The string is empty if the filename hasn't been set.

(This bit of code was introduced in #717.)


happycube:
oops, jumped the gun a bit, hopefully i didn't break anything ;)

PR #730, opened 2022-05-28
Test suite improvements

atsampson:
Several changes to the test suite:

- Make it possible to run the tests locally with `make check`.
- Skip the decode tests if `testdata` isn't available.
- Write the ld-cut output file to the `testout` directory.
- Add unit tests for `JsonReader`, and fix a couple of bugs they revealed.

It would be nice to have the CI run the tests both for Qt 5 and for Qt 6, but I'm not sure all the Qt 6 deps are available...


happycube (approved):
I don't have the energy to really understand them, but they sound good :)

simoninns (approved):

PR #734, opened 2022-05-29
Fix some compiler warnings

atsampson:
Some of QScopedPointer's API is deprecated in favour of std::unique_ptr in Qt 6.1; use std::unique_ptr instead.

Avoid a signed/unsigned comparison in ld-ldf-reader. **Edit:** with CI, this revealed a fault in the seeking logic - fixed.

Remove some dead code in ld-chroma-encoder. (I had to do some testing to convince myself it really wasn't needed; separate PR coming to add the test.)

### DIFF
--- a/ld-ldf-reader.c
+++ b/ld-ldf-reader.c
@@ -2,7 +2,7 @@
  * Decode 16-bit samples from a compressed file using ffmpeg's libraries.
  *
  * Copyright (c) 2019-2021 Chad Page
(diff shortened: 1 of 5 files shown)


simoninns:
Failing CI @atsampson - please verify

atsampson:
This turned out to be another bug that fixing the comparison revealed - so I've included a fix for that as well...

PR #743, opened 2022-06-04
Fix discmap.cpp warning and don't use pyfftw on windows

oyvindln:
Seems pyfftw and python threads don't mix so disable on windows where we have to use Thread.
It doesn't play super well with Process either but at least it doesn't completely break.

Could potentially make the discmap.cpp thing a bit cleaner but wasn't sure if the constructor for Frame was used elsewhere.

### DIFF
--- a/ld-decode
+++ b/ld-decode
@@ -4,8 +4,6 @@ import sys
 import argparse
 import traceback
 
-from multiprocessing import Process, Pipe
-
 from lddecode.core import *
 from lddecode.utils import *
 from lddecode.utils_logging import init_logging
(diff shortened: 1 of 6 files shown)


atsampson:
Looks good to me. It would be worth checking if PyFFTW is still faster than numpy's own FFT - I think the numpy code has improved a bit since I added the PyFFTW code...

oyvindln:
Oh, added another thing I almost forgot - a parameter to skip the phaseID check so I can turn that off for tape sources if needed. Default behaviour isn't changed

PR #753, opened 2022-06-08
Add LineNumber and rework how line numbers are shown in the scope

atsampson:
The LineNumber class represents a line number within a ComponentFrame/OutputFrame, and can convert between the various different representations of line numbers that the ld-decode tools use - plus the standard way of numbering PAL/NTSC lines. There are various places in the code that could potentially use this in the future, but for now...

The oscilloscope view has been converted to use LineNumber, so it now shows the selected X/Y coordinates within the frame, and the line numbers that you need if you're referring to a standard or working with the metadata:
scope3

Thanks to various people on IRC/Discord for workshopping the UI changes!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,7 @@ clean-tools:
 
 TESTCASES = \
 	check-library-filter \
+	check-library-linenumber \
 	check-library-metadata \
 	check-library-vbidecoder \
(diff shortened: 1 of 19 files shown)


atsampson:
Added a function to parse VideoSystem names, since #752 wanted to do that.

PR #754, opened 2022-06-08
Add NTSC-J support to ld-chroma-encoder

ifb:
Adds '--no-setup' to disable adding setup.

On the decoder side, all this really does is change black16bIre in the JSON.  test-chroma.sh will need some refactoring if we care to test these various NTSC-specific encoder options (no-setup and the wideband-yuv/wideband-yiq modes).


atsampson:
Looks good to me!

PR #760, opened 2022-06-22
NTSC chroma-encoder filter changes

ifb:
Add `--chroma-mode narrowband-q`

This low-passes the Q channel such that it uses about half the bandwidth of the I channel.  Non-equiband Y'IQ encoder/decoders were never widespread because of the extra complexity according to Poynton.  Including it here for testing/completeness.

NTSC uses a different chroma low-pass than PAL for the wideband channels.  Use the filter suggested by Clarke.

While neither of these two filters from Clarke quite match the filter template, anything I created in Iowa Hills that did meet the spec was more complex and ended up lowering PSNR vs. Clarke.  DSP gurus are welcome to revisit this.


atsampson:
Looks good to me - thanks!

As you've been thinking about NTSC I/Q filters, do you have any suggestions for the ones in the NTSC decoder (`colorlpi` and `colorlpq` [generated here](https://github.com/happycube/ld-decode/blob/master/scripts/filtermaker#L203) and [used here](https://github.com/happycube/ld-decode/blob/master/tools/ld-chroma-decoder/comb.cpp#L647))? The existing ones are Butterworth filters which aren't phase-linear, and this is quite visible for the narrow Q filter in particular.

PR #761, opened 2022-06-24
Workaround for weird linker issue with connect/qtw on windows

oyvindln:
Using connect with scaleDivChanged gives a weird linker error on windows for some reason. I could not figure out why, but using the old-style connect that operates on ```const char*``` works fine.

Add math define to make M_PI available on windows (M_PI is not-standard but a common extension, there is no built-in pi constant in std c++ headers before c++20)

also bump pro files to c++17. C++11 is ancient, c++17 is the default on the system we support I believe.

### DIFF
--- a/tools/ld-analyse/ld-analyse.pro
+++ b/tools/ld-analyse/ld-analyse.pro
@@ -160,7 +160,7 @@ unix:!macx {
 win32: {
     LIBS += qwt.lib
 }
-win32:INCLUDEPATH += "D:\prog\vcpkg\installed\x64-windows\include\qwt"
+
 macx {
 INCLUDEPATH += "/usr/local/lib/qwt.framework/Versions/6/Headers"
 LIBS += -F"/usr/local/lib" -framework qwt
--- a/tools/ld-analyse/ld-analyse.pro
+++ b/tools/ld-analyse/ld-analyse.pro
@@ -16,13 +16,14 @@ TEMPLATE = app
 # depend on your compiler). Please consult the documentation of the
 # deprecated API in order to know how to port your code away from it.
 DEFINES += QT_DEPRECATED_WARNINGS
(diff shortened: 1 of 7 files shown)


tools/ld-analyse/ld-analyse.pro
```diff
@@ -159,6 +160,7 @@ unix:!macx {
 win32: {
     LIBS += qwt.lib
 }
+win32:INCLUDEPATH += "D:\prog\vcpkg\installed\x64-windows\include\qwt"
```
atsampson, line 163: Is that addition intentional? I assume that's a path on your machine...

atsampson:
This looks basically OK to me. If we're requiring C++17, then there are some other cleanups that I can do as well (removing some C++14 workarounds, and I think there were a couple of places that could use `make_unique`).

With the connect calls, could you put those in a conditional for Windows, so other OSs don't lose the benefit of type-safety?

tools/ld-analyse/ld-analyse.pro
oyvindln: Ah yeah that  just me forgetting to remove it

oyvindln:
Okay, can make it conditional and leave the type safe connects in for non-windows.

Are there any possible systems we still support where C++17 support would be missing? Even Ubuntu 18.04 has GCC 7.3 (and optionallly gcc 8) which supports most of c++17 other than a few lesser used bits, though I think we might require 20.04 as minimum now anyhow which has GCC 9.3 so it shouldn't be any issues using that as the default and not needing workarounds for older ones.

atsampson:
I don't think so - the CI tests use Ubuntu 20.04, which is the oldest version we support, and it's building OK there.

atsampson:
Looks good. I'll merge these by hand along with removing the pre-C++17 workarounds.

PR #772, opened 2022-08-19
Add DTS support to ld-process-efm

### LINKED ISSUE #527: Support DTS Processing / Extraction
I just captured a minute of a DTS laserdisc. I was thinking as this is just the PCM track filled with DTS it might work, if I turn off error concealment but no. The issue starts quite early as all F3 frames are discarded. It won't find sync0. But the fix was very easy. Change the sync0 bit to 0x812 and that's it - DTS works. Yeah, you won't have error concealment and I'm not sure if the statistics are correct but I'd say it's a start.

I changed f3frame.cpp line from 136 to
```
    if (efmValues[0] == 0x812) {
        // Sync 0
        subcodeSymbol = 0;
        isSync0 = true;
```
Actually it's just replacing 0x801 with 0x812 in 136. Of course this will break regular PCM decoding it was just a quick hack to see if it works.

I can provide a lds or tbc sample to work with but note that the capture quality on the one I just did is far from perfect. I'm still in the process of setting everything up.

atsampson:
This is based on @DreckSoft's work in #527. It adds a `--dts` option to accept the alternative Sync 0 value that some DTS LaserDiscs use. Since audio error concealment doesn't make sense for DTS, it also changes the default error treatment when the option is used (and checks that you haven't specified multiple treatment options).

It's been tested on clips from the following [DTS discs](https://www.lddb.com/list.php?format=ld&list=dts):
- Armageddon - decodes as normal
- Goldeneye - needs `--dts --time`
- Jurassic Park - needs `--dts --time`
- Jurassic Park: The Lost World - decodes as normal

Fixes #527.

Note to self: will need to add the new option, and some advice on using it, to the wiki.

### DIFF
--- a/tools/ld-process-efm/Datatypes/f3frame.cpp
+++ b/tools/ld-process-efm/Datatypes/f3frame.cpp
@@ -47,7 +47,7 @@ F3Frame::F3Frame()
     }
 }
 
-F3Frame::F3Frame(uchar *tValuesIn, qint32 tLength)
+F3Frame::F3Frame(uchar *tValuesIn, qint32 tLength, bool audioIsDts)
 {
     validEfmSymbols = 0;
     invalidEfmSymbols = 0;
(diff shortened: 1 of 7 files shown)


atsampson:
I've updated the [ld-process-efm](https://github.com/happycube/ld-decode/wiki/ld-process-efm) wiki page to match this.

PR #776, opened 2022-08-20
Remove the mode option from the Transform decoders

atsampson:
There doesn't appear to be any reason to use level mode (the BBC hardware decoders use threshold mode), so removing the option simplifies both the code and the interface.

Note to self: update the wiki.


happycube:
This change is conflicting with one of the last few ones, needs updating before I can commit.

atsampson:
Rebased - should be good now.

atsampson:
Updated the ld-analyse and ld-chroma-decoder pages on the wikl for this and the other recent changes.

PR #780, opened 2022-08-26
First CMake files

### LINKED ISSUE #769: Support for more Linux distros
This will be a living document, updated as I try installing it on various Linux distros.

Here is much more detail: https://docs.google.com/document/d/132ycIMMNvdKvrNZSzbckXVEPQVLTnH_YX0Oh3lqtkkQ/edit?usp=drivesdk

lizardgai4:
Fixes #769 

Plus new gitignore parameters for CMake-generated files,  It's a good start.

Current flaws:
- The program only works on Linux (but porting is very hard)

Fixed flaws:
- Requires code modifications to work on multiple Linux distros
- ~~-ld-analyse requires modifications to the generated Makefile to compile~~ Fixed (on Ubuntu at least)
- FindQWT code and ld-analyse code currently occupy the same CMakeLists.txt file.
- Having Qt6 installed at all might break installation (explicitly mentioning Qt5 instead of just Qt in the CMake files may fix it)
- library install locations are hard-coded to /usr/lib64
- ld-analyse compiles and runs on Ubuntu, but visual assets (png and svg files) currently don't install with the program.


happycube:
I'm still on Ubutu 20.04, this doesn't work for me:

"
-- ---------Beginning cmake build of ld-ldf-reader---------
CMake Error at CMakeLists.txt:3 (cmake_minimum_required):
  CMake 3.22 or higher is required.  You are running version 3.16.3
"

CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+set(CMAKE_CXX_STANDARD 17)
+
+set(CMAKE_AUTOUIC_SEARCH_PATHS "${SMAKE_CURRENT_SOURCE_DIR}/ui")
+
+# Include all the paths
+message( STATUS "--------- Finding packages---------")
+
+find_package(Qt5 REQUIRED COMPONENTS Core)
```
atsampson, line 15: I'd definitely get it working with Qt 5 first, but we will need Qt 6 support as well (see the DomesdayDuplicator commits for how we handled it there without needing to make too much stuff conditional).

CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+add_subdirectory(tools/ld-process-vbi)
+add_subdirectory(tools/ld-process-vits)
+
+add_subdirectory(tools/ld-analyse)
+
+# ld-ldf-reader executable
+add_executable(${PROJECT_NAME} "ld-ldf-reader.c" )
+
+target_link_libraries(${PROJECT_NAME} PkgConfig::LIBAV )
+target_link_libraries(${PROJECT_NAME} ld-tbc)
```
atsampson, line 48: ld-ldf-reader doesn't need tools/library - it's a hacked version of one of the ffmpeg examples so it only needs ffmpeg.

CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+install(TARGETS ${PROJECT_NAME} DESTINATION "")
+
+set(OUTPUT ${CMAKE_CURRENT_BINARY_DIR})
+set(DEPS "${CMAKE_CURRENT_SOURCE_DIR}/lddecode/__init__.py")
+
+# install Python files
+
+message( STATUS "--------- Installing Python files to ${OUTPUT} ---------")
+
+find_program(PYTHON NAMES python python3)
```
atsampson, line 60: This would need to look for python3 first, since python is usually Python 2, but it would be better to use the `Python3` CMake module:

```
find_package(Python3 3.6 REQUIRED COMPONENTS Interpreter)
```

(I'm not sure what the minimum version of Python we require is - it's at least 3.6 for f-strings.)

CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+
+add_custom_command(
```
atsampson, line 62: There's got to be a better way of building and installing Python modules from CMake then running setup.py...?

tools/ld-analyse/CMakeLists.txt
```diff
@@ -0,0 +1,246 @@
+message( STATUS "---------Beginning cmake build of ld-analyse---------")
+
+# Qt Widgets for Technical Applications
```
atsampson, line 3: Put this in `cmake/FindQwt.cmake` at the top level - don't copy the text in yourself.

tools/ld-analyse/CMakeLists.txt
```diff
@@ -0,0 +1,246 @@
+message( STATUS "---------Beginning cmake build of ld-analyse---------")
+
```
atsampson, line 2: The subdirs all check for dependencies and the Git version individually, so there's a lot of duplication. It would be better to check for all the deps in the top-level `CMakeLists.txt` (as GNURadio etc. do) - then you can also enable/disable subdirs if dependencies are missing.

CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+# ld-ldf-reader executable
+add_executable(${PROJECT_NAME} "ld-ldf-reader.c" )
+
+target_link_libraries(${PROJECT_NAME} PkgConfig::LIBAV )
+target_link_libraries(${PROJECT_NAME} ld-tbc)
+target_link_libraries(${PROJECT_NAME} ld-filter)
+
+install(TARGETS ${PROJECT_NAME} DESTINATION "")
+
+set(OUTPUT ${CMAKE_CURRENT_BINARY_DIR})
```
atsampson, line 53: You need to `include(GNUInstallDirs)` to get this definition (it's being pulled in by a subdir at the moment - but see comment below about moving tests like this to the toplevel anyway).

tools/ld-analyse/CMakeLists.txt
```diff
@@ -0,0 +1,246 @@
...
+set(CMAKE_AUTOMOC ON)
+include(GNUInstallDirs)
+set(CMAKE_AUTOUIC ON)
+
+find_library(QWT_LIBRARY_RELEASE NAMES qwt qwt-qt5 qt5-qwt PATH_SUFFIXES lib)
+find_package(QT NAMES Qt5 REQUIRED COMPONENTS Core Gui Widgets HINTS ${QT_LIBRARY_DIR})
+find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Core Gui Widgets)
+
+# Name all the files
+add_executable(ld-analyse MACOSX_BUNDLE
```
atsampson, line 128: Is it necessary to specify the .h files here? (It may well be for Qt's autogeneration stuff to work - it's just that normally CMake can figure out the headers for itself.)

tools/ld-analyse/CMakeLists.txt
```diff
@@ -0,0 +1,246 @@
...
+    ../ld-chroma-decoder/palcolour.cpp ../ld-chroma-decoder/palcolour.h
+    ../ld-chroma-decoder/comb.cpp ../ld-chroma-decoder/comb.h
+    ../ld-chroma-decoder/componentframe.cpp ../ld-chroma-decoder/componentframe.h
+    ../ld-chroma-decoder/outputwriter.cpp ../ld-chroma-decoder/outputwriter.h
+    ../ld-chroma-decoder/transformpal.cpp ../ld-chroma-decoder/transformpal.h
+    ../ld-chroma-decoder/transformpal2d.cpp ../ld-chroma-decoder/transformpal2d.h
+    ../ld-chroma-decoder/transformpal3d.cpp ../ld-chroma-decoder/transformpal3d.h
+    ../ld-chroma-decoder/framecanvas.cpp ../ld-chroma-decoder/framecanvas.h
+    ../ld-chroma-decoder/sourcefield.cpp ../ld-chroma-decoder/sourcefield.h
+    ../library/filter/firfilter.h
```
atsampson, line 152: The stuff you've built as libraries shouldn't be listed here - you will get these by linking with the library.

tools/ld-analyse/CMakeLists.txt
```diff
@@ -0,0 +1,246 @@
...
+# Get the latest abbreviated commit hash of the working branch
+execute_process(
+	COMMAND git log -1 --format=%h
+	WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+	OUTPUT_VARIABLE GIT_COMMIT_HASH
+	OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+# Pass it on to the source code
+add_definitions(-DAPP_BRANCH="\""+${GIT_BRANCH}+"\"" -DAPP_COMMIT="\""+${GIT_COMMIT_HASH}+"\"")
```
atsampson, line 182: This is deprecated - use [add_compile_definitions](https://cmake.org/cmake/help/latest/command/add_compile_definitions.html) which will make the quoting a bit less horrible.

lizardgai4:
@happycube is it possible to upgrade to CMake 3.22 on your machine?

tools/ld-analyse/CMakeLists.txt
```diff
@@ -0,0 +1,246 @@
...
+endif()
+
+#install(TARGETS ld-analyse
+#    BUNDLE DESTINATION .
+#    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+#)
+
+install(TARGETS ${PROJECT_NAME} DESTINATION "")
+
+# Consider using qt_generate_deploy_app_script() for app deployment if
```
atsampson, line 244: Remove this comment (from the converter script).

tools/ld-chroma-decoder/CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+
+# ld-chroma-decoder executable
+
+add_executable(${PROJECT_NAME} ${SOURCES} ${HEADERS_MOC})
+
+target_link_libraries(${PROJECT_NAME} PRIVATE Qt5::Core fftw3 ld-tbc ld-filter)
+
+install(TARGETS ${PROJECT_NAME} DESTINATION "")
+
+# Make a library for ld-analyse to use
```
atsampson, line 100: Same comments as the other library below.

tools/ld-chroma-decoder/CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+
+add_library(ld-chroma-lib SHARED
+	"comb.h"
+	"palcolour.h"
+)
+
+target_include_directories(ld-chroma-lib PUBLIC ./)
+
+set_target_properties(ld-chroma-lib PROPERTIES LINKER_LANGUAGE CXX)
+install(TARGETS ld-chroma-lib LIBRARY DESTINATION /usr/lib64 ARCHIVE DESTINATION /usr/lib64)
```
atsampson, line 114: Don't hardcode any installation paths - use the variables provided by `GNUInstallDirs`. (`/usr/lib64` is only correct for some architectures, on some distributions - e.g. it's wrong on Debian/Ubuntu - but more generally users should be able to specify their own installation prefix for local installation.)

tools/ld-chroma-decoder/encoder/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
...
+
+ADD_DEFINITIONS(${QT_DEFINITIONS})
+
+QT_WRAP_CPP(HEADERS_MOC ${HEADERS})
+
+# ld-chroma-encoder executable
+
+add_executable(${PROJECT_NAME} ${SOURCES} ${HEADERS_MOC})
+
+target_link_libraries(${PROJECT_NAME} PRIVATE Qt5::Core fftw3 ld-tbc ld-filter)
```
atsampson, line 65: The only tools that need FFTW are ld-analyse and ld-chroma-decoder - it's used by the Transform PAL decoder.

tools/ld-disc-stacker/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+add_definitions(-DAPP_BRANCH="\""+${GIT_BRANCH}+"\"" -DAPP_COMMIT="\""+${GIT_COMMIT_HASH}+"\"")
+
+find_package(Qt5 REQUIRED COMPONENTS Core Widgets Gui)
+
+ADD_DEFINITIONS(${QT_DEFINITIONS})
+
+QT_WRAP_CPP(HEADERS_MOC ${HEADERS})
+
+find_library(QWT_LIBRARY_RELEASE NAMES qwt qwt-qt5 qt5-qwt PATH_SUFFIXES lib)
+find_package(QT NAMES Qt5 Qt5 qwt-qt6 REQUIRED COMPONENTS Core Gui Widgets HINTS ${QT_LIBRARY_DIR})
```
atsampson, line 55: I think you meant to include qwt-qt6 in the line above (but you don't want it for Qt 5-only-for-now anyway).

tools/library/filter/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+message( STATUS "---------Beginning cmake build of ld-filter---------")
```
atsampson, line 1: I would build all of `tools/library` as a single llibrary, and it should be an internal static library, not an installed shared library. This is for two reasons - we don't want to have to version and maintain a shared library interface, and there's lots of performance-critical stuff in the library that can benefit from LTO.

tools/library/filter/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+project( ld-filter )
+
+#add_compile_options(-std=c++17
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread -O3 -Wall -ftree-vectorize -ffast-math -funroll-loops")
```
atsampson, line 5: It's not a good idea to hardcode compiler flags like this, but especially don't build ld-decode with `-ffast-math` - that makes maths operations faster at the cost of giving less accurate results!

atsampson (commented):
This isn't a detailed review, but a few comments from a first look through this...

The GitHub CI config needs updating to build the project with CMake - you should be able to copy the one from DomesdayDuplicator, which builds on both the oldest and newest versions of Ubuntu (do it with Qt 5 only for now).

The test rules from the top-level Makefile need converting to CMake tests - this should be easy enough since they're all just running a command and checking it exits 0.

lizardgai4:
@atsampson I will make those changes tonight and test them before committing

tools/ld-analyse/CMakeLists.txt
lizardgai4: I based this file off of [the top-level CMakeLists.txt file of DomesdayDuplicator](https://github.com/oyvindln/DomesdayDuplicator/blob/master/Linux-Application/DomesdayDuplicator/CMakeLists.txt), which is also designed for GUIs

tools/ld-analyse/CMakeLists.txt
lizardgai4: - **Libraries and filenames:** Compiles and runs
- **Libraries only:** Undefined references to functions when attempting to compile
- **Filenames only:** Cannot compile because it can't find certain files

TL;DR both libraries and filenames are required to compile

lizardgai4:
After some collaboration, it works with nearly every Linux distro I tested, no source code tweaking required!

Working:
- **Ubuntu** 22.04 LTS
- **Fedora** 36
- **Arch BTW** (rolling release)
- **Manjaro** 21.3.7 Ruah
- **OpenSUSE** Leap 15.4
- **Solus** 4.3

Not working (yet): **Slackware** 15.0

PR #787, opened 2022-08-28
Fix chroma rotation for phase comp decoder and make some json values that are integers actually output as int

oyvindln:
I added a 10 degree offset to the chroma rotation in the phase compensating decoder as that seemed to fix phase offset, but after some digging it seems it was the samples that were off so it now correctly rotates 33 degrees which should give the correct result if the burst/color relationship in the sample is correct.

Also made some values that are supposed to be integers actually be output as integers to the json, round didn't change the datatype apparently so it was still outputting .0 int the json for those values.

### DIFF
--- a/lddecode/core.py
+++ b/lddecode/core.py
@@ -3776,7 +3776,7 @@ class LDdecode:
             "syncConf": f.compute_syncconf(),
             "seqNo": len(self.fieldinfo) + 1,
             "diskLoc": np.round((self.fieldloc / self.bytes_per_field) * 10) / 10,
-            "fileLoc": np.floor(self.fieldloc),
+            "fileLoc": int(np.floor(self.fieldloc)),
             "medianBurstIRE": roundfloat(f.burstmedian),
         }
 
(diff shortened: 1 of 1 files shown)


happycube (commented):
Is the -10 degree comment in comb.cpp still needed?

oyvindln:
No can remove that, forgot that it was there.

happycube:
@atsampson Does the encoder need correction here or is something wrong with this patch set?

> from PIL import Image
> 
> img1 = Image.open('/home/cpage/ld-decode/git/testout/test.output-pc-ntsc2d-rgb.png')
> img2 = Image.open('/home/cpage/ld-decode/i787//testout/test.output-pc-ntsc2d-rgb.png')
> 
> for i in [0, 110, 220, 330, 440, 550, 660, 700]:
>     print(i, img1.getpixel((i,200)), img2.getpixel((i,200)))

> 0 (189, 189, 189) (189, 189, 189)
> 110 (174, 174, 97) (167, 178, 95)
> 220 (82, 159, 45) (75, 160, 59)
> 330 (0, 173, 97) (0, 165, 135)
> 440 (62, 99, 62) (60, 99, 69)
> 550 (173, 0, 97) (176, 0, 59)
> 660 (82, 0, 46) (84, 0, 23)
> 700 (0, 0, 191) (16, 0, 195)

oyvindln:
The rotation part was fixed in @atsampson 's PR, but the json stuff still needs merging.

happycube:
Can you update your side to include the rotation fix?  Once that's done I'll merge it.

oyvindln:
ok should be fixed now

PR #789, opened 2022-08-29
Fix chroma encoding in NTSCEncoder

atsampson:
This should fix the encoding problem identified in the discussion for #787 - I ended up reworking how it figures out the phase offset for each sample, and fixing how it outputs phaseID to the metadata. This includes @oyvindln's change to Comb from #787 so the tests will pass (but the other change in #787 should rebase cleanly on top of this).

(Note that the output is still offset to the right when compared with ld-decode's output for both PAL and NTSC - this is because it's following SMPTE 244M/EBU Tech 3280 by having each line contain all the blanking samples, then all the active samples, whereas ld-decode starts each line with the sample immediately before 0H.)

CC @oyvindln and @ifb.


ifb:
Looks good to me.  Thanks!

PR #795, opened 2022-09-11
Optimize downscale including using numba

oyvindln:
Informal test using timeit indicates it's like 100x faster (doing loops in python directly is VERY inefficient). It was one of the functions identified via the profiler.

Should give a small speedup on samples with analog audio.

was a little unsure about what some of the arrays represent so the function docs needs looking over

Also, noted that the 16-bit PCM values are limited to +/- 32766 rather than the typical -32,768 to 32,767, wasn't sure if this is intentional and maybe required by the format or if it's an oversight.

### DIFF
--- a/lddecode/core.py
+++ b/lddecode/core.py
@@ -17,6 +17,8 @@ if platform.system() != "Windows":
 import numpy as np
 import scipy.signal as sps
 import scipy.interpolate as spi
+import numba
+from numba import njit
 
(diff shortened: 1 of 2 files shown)


oyvindln:
should be downscale_audio, haven't touced downscale

happycube:
I think it's intentional so that dropouts could be assigned to one of the boundaries.

PR #806, opened 2022-10-17
Fix building with cmake on windows

oyvindln:
Neither qwt nor pkg-config seems to include a .pc file on windows, so this borrows a findQwt script from [QGis](https://github.com/qgis/QGIS/blob/master/cmake/FindQwt.cmake) to make handling it easier.

The other addition needed was to define ```_USE_MATH_DEFINES``` as ```M_PI``` is a non-standard extension and as such not available from ```<cmath>``` when compiling with MSVC otherwise. 

Now builds with cmake+msbuild on windows if the needed packages are installed via vcpkg and using the ms dev environment command line and vcpkg cmake toolchain file. 

ld-ldf-reader is still disabled for windows as it uses some linux/unix-specific functions, also disable it for macos as @kaliohelix on discord reported it didn't build properly there either.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ set(USE_QT_VERSION "" CACHE STRING
 )
 
 # Not working on Windows and macos as of yet. The decoders are able to use ffmpeg binaries instead if needed.
-if(LINUX)
+if((NOT WIN32) AND (NOT APPLE))
     option(BUILD_LDF_READER
         "build ld_ldf_reader"
         ON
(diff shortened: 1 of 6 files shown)


oyvindln:
Ok, it should now try with pkg-config first.

Which distros does it fail on? It's a different find script than what was used last previously.

It looks like qwt explicitly disables pkg-config generation on non-UNIX systems:
https://sourceforge.net/p/qwt/git/ci/develop/tree/qwtconfig.pri
Also Qt 6.0 and 6.1 lacked pkg-config. It's possible the config to enable it for qwt works on win but I don't know.

oyvindln:
Also changed the conditional for ld-ldf-reader, seems the LINUX variable was added only in the last version.

oyvindln:
bump @atsampson

PR #823, opened 2022-12-28
use with self.lock, exit loader properly, demodblock rawdata to input 

oyvindln:
Use "with" with lock to ensure proper cleanup and make it a bit cleaner
Make the ffmpeg loader exit properly (tends to end with some "broken pipe" warning otherwise)
supply a reference to raw data in demodblock call (not used in ld-decode but is used particularly with cvbs-decode, helps avoid having to re-create it from the fft)


happycube (approved):
Looks very good to me, thanks!

PR #824, opened 2022-12-30
Add VITC support to ld-process-vbi

### LINKED ISSUE #474: Add VITC support to ld-process-vbi
Some PAL discs include [vertical interval timecode](https://tech.ebu.ch/publications/tech3097), presumably as a by-product of the mastering process. 

louvre1-leadin

ld-process-vbi could decode this and store it in the JSON.

atsampson:
The main thing here is support for extracting VITC in ld-process-vbi. I've tested this with several PAL discs and it seems to works well, but I don't have any NTSC or videotape samples - so it might need some tweaking after further testing.

In addition, I've refactored ld-process-vbi to make it easier to add new formats, and cleaned up several other warnings and oddities spotted along the way.

Fixes #474.


atsampson:
Thanks to Brad and Harry on IRC, I've now tested this with some NTSC VHS samples as well, and it appears to work happily for those as well. I've added one more change to the above - ld-process-vbi was setting vbi.inUse even if there weren't any valid VBI lines, so the JSON will now be a bit smaller for non-LaserDisc output.

atsampson:
... and now with a `VitcDecoder` class that works like `VbiDecoder`, so it can show the decoded timecode (along with the VBI position, for good measure) in ld-analyse:

status

happycube:
Looks good for my bits, but I'll leave it to Simon to approve

atsampson:
A couple more tweaks that should make it work for PAL-M (which also makes me wonder if the CC decoder should be enabled for PAL-M?).

PR #827, opened 2023-01-03
Handle more variation in VITC

atsampson:
@harrypm's Eden Project sample failed to decode for a couple of reasons: it was using VITC lines we weren't looking at, and the time between the sync pulse and VITC start meant the end of VITC was (just) outside the active region.

Fix this by looking at all the possible lines, and giving the decoders the complete composite line to look at rather than just the active region.


atsampson:
Reworked after testing with some commercial tapes - The Matrix VHS (S016985) has VITC with timing that's out of spec in the other direction (in fact, it starts in spec, then jumps a bit's width to the left as the film starts). So allow any start position after the colourburst.

PR #831, opened 2023-01-12
readd some stuff that got removed in a merge and some build fixes

oyvindln:
One of the last prs seems to have removed the code closing the loader properly on exit and supplying a reference to raw data to demodblock so readding that

Move the code in ```ld-decode``` to a ```main.py``` and make the original a shim calling it to make it easier to make a single file call any of ld-decode/vhs-decode etc and also modern python project seems to generally use some config file defiined entry point for callable files in a src directory so that would make that easier too. 

Change cmake file to skip looking for qt gui bits if they are not used so one can build the cli tools without that available and only supply ```--prefix``` to setup.py install if it's manually set. If --prefix is supplied even if it's the default /usr/local one seems to make the installed files not find the modules by default without adding some stuff to pythonpath.


CMakeLists.txt
```diff
@@ -154,6 +159,18 @@ if(BUILD_PYTHON)
 
     set(PYTHON_BUILD_DIR ${CMAKE_CURRENT_BINARY_DIR}/python-build)
 
+    # python setup.py install seems to behave differently if prefix
+    # is supplied, requiring setting pythonpath after install.
+    # If a custom prefix is not set, don't supply prefix at allocate_lock
+    # /usr/local is the default on linux.
+    # Install not really tested on windows as of yet.
+    if(${CMAKE_INSTALL_PREFIX} STREQUAL "/usr/local")
+        set(PYTHON_PREFIX_LINE, "")
```
atsampson, line 168: I don't think that comma's meant to be there - otherwise looks good.

oyvindln: Fixed

happycube (approved):

PR #833, opened 2023-01-15
Add Closed Caption support for PAL

atsampson:
We haven't found any examples of PAL CC on LaserDisc yet, but it certainly exists on commercial VHS tapes - the format is the same, with 32 bit times per line, but it's on field line 22 rather than 21.

This will need an update to the JSON spec once merged, since it moves `ntsc.ccData[01]` to `cc.data[01]`. The parser understands the old location as well when reading existing JSON files.

### DIFF
--- a/tools/ld-analyse/closedcaptionsdialog.ui
+++ b/tools/ld-analyse/closedcaptionsdialog.ui
@@ -17,7 +17,7 @@
    </size>
   </property>
   <property name="windowTitle">
-   <string>NTSC Closed Captions</string>
+   <string>Closed Captions</string>
   </property>
   <property name="windowIcon">
    <iconset resource="ld-analyse-resources.qrc">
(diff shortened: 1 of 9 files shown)


happycube (approved):

ifb:
Closed Captions (CC3/CC4 + XDS) are also on line 284 in 525-line systems (not sure about 625, but probably the next line down in field 2).

Obviously it doesn't need to be implemented now (or ever), but something to think about.  It would be more interesting for non-LD formats like VHS that are more likely to have second language (CC3) or XDS (V-Chip) present.

You can get time of day and station IDs from broadcast recordings with XDS.

atsampson:
ld-process-vbi works per field, so it'll pick up both line 21 and line 284 (i.e. line 21 of the second field). I don't know if the SCC exporter in ld-export-metadata correctly handles anything other than CC1 at the moment, though - ld-analyse definitely won't, because (as with LaserDisc VBI) it merges the CC from both fields together.

PR #838, opened 2023-01-26
Add fftw to ld-analyse include dirs

oyvindln:
Not sure why missing this didn't cause errors in the windows build before but hit it now. fftw is used by something in ld-analyse so if it's not included it won't be found if it's not in a standard location like on linux.


atsampson:
Looks good to me. (The "something" is the Transform PAL decoders - they use FFTW to do the FFTs.)

PR #845, opened 2023-03-01
Make compatible with numpy 1.24

Tatsh:
Related to https://github.com/oyvindln/vhs-decode/pull/78


atsampson:
Looks good to me! I can't test with numpy 1.24 because there's no numba release compatible with it yet, but these aliases have been deprecated since numpy 1.20 and it'll get rid of the deprecation warnings.

PR #846, opened 2023-03-05
y4m: Always signal field dominance correctly

ifb:
Field dominance can change depending on the first active line and any padding at the top.

This has been potentially broken ever since dynamic padding and the ffll/lfll/ffrl/lfrl options were added to `ld-chroma-decoder`


atsampson:
Looks good to me - thanks!

PR #847, opened 2023-03-24
ld-process-vbi: Support reading VIDEO ID (IEC 61880) data in the VBI

ifb:
Lines 20 and 283 can contain information about aspect ratio, CGMS-A, and APS.  Store these 14-bits in the JSON.

NTSC only.  PAL uses WSS to provide similar functionality.

### DIFF
--- a/tools/ld-process-vbi/CMakeLists.txt
+++ b/tools/ld-process-vbi/CMakeLists.txt
@@ -5,6 +5,7 @@ add_executable(ld-process-vbi
     main.cpp
     fmcode.cpp
     vbilinedecoder.cpp
+    videoid.cpp
     vitccode.cpp
     whiteflag.cpp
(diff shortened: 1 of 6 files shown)


ifb:
I haven't looked at writing any tests, but can work on that.

Yet to be decided is if and how to display the information in ld-analyse:

aspect ratio
- 4:3
- 16:9
- Letterbox

CGMS-A
- Copying is permitted without restriction
- Condition not to be used
- One generation of copies may be made
- No copying is permitted

APS trigger bits
- PSP off
- PSP on, split burst off
- PSP on, 2-line split burst on
- PSP on, 4-line split burst on

Of the 14-bit payload, only 6 bits are defined in IEC 61880.  It's possible that other data could be inserted as well?

PR #849, opened 2023-04-01
gen_chroma_vid.sh CVBS support

peppi001:
Please add CVBS option to gen_chroma_vid.sh.
THX


harrypm:
I think your getting VHS-Decode mixed up with LD-Decode which is composite to begin with and like with CVBS decode produces a single TBC file that does not need scripting as it can be directly used with the chroma decoder in a single piped FFmpeg command as noted across both wikis.

peppi001:
Yes, you're right. Sorry I posted to the wrong place.
Sorry :(

PR #850, opened 2023-04-06
Change system python path to env python path

pepie34:
Don't use the system python path but the current python env path so the building process can be done on custom python environments.

This is needed for example to compile the project on macOS under conda python env. 

/usr/bin/env python3 resolves to /usr/bin/python3 when no custom python env is activated

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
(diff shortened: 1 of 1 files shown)


atsampson:
Looks OK to me. I've just had a grep through all the Python 3 modules I package for a distribution, and it's about equally common to have a #! line using env and to not have a #! line at all.

PR #853, opened 2023-04-25
change icon ld-analyse + install

vrunk11:
Change icon of ld-analyse

add icon to the executable on windows , add .desktop and icon on linux during the make install
adjust small icon in the status bar to be 16x16 to avoid downscaling


tools/ld-analyse/CMakeLists.txt
```diff
@@ -43,7 +52,22 @@ if(WIN32)
...
 endif()
 
-install(
-        TARGETS ld-analyse
-        BUNDLE DESTINATION bin
-)
+if(UNIX)
+	install(TARGETS ld-analyse BUNDLE DESTINATION bin)
+	#add .tbc mime type
+	install(CODE "execute_process(COMMAND xdg-mime install --novendor --mode system ${CMAKE_CURRENT_SOURCE_DIR}/install/ld-analyse.xml)")
```
atsampson, line 58: Running commands like this at installation time won't work - you need to respect CMAKE_INSTALL_PREFIX and DESTDIR for people who're building packages or installing into a custom location. And xdg-mime is adjusting the current user's settings, not installing a system default - you want to install a .xml file for the MIME type.

You could see how Marble does it for one example - [data](https://invent.kde.org/education/marble/-/blob/master/data/CMakeLists.txt) and [application](https://invent.kde.org/education/marble/-/blob/master/src/apps/marble-qt/CMakeLists.txt).

PR #874, opened 2023-12-02
Fix random bits of CLV popping up in a CAV frame

happycube:
rktk gave me a sample that threw a (non-fatal) exception in the decoding code because of a CLV fragment.  Fixed that and then also replaced "CAV Frame #None Unknown" with "CAV Pulldown/Telecine Frame"


happycube:
Added 874.ldf to ld-decode-testdata, which has the previously-errored frame in the second frame.

PR #875, opened 2023-12-11
Add stdin support for ld-decode

Inky1003:
A simple change that would help me a lot.
As ld-decode has no progress bar or ETA,
I can't know when the decodings would end.

Then, If I pipe the RF capture from pv into
x-decode, It works, and I only need to redirect
the stdout and stderr to null. :)

It does not affect the common file opening.


happycube:
I could see cases where this doesn't work (but the demodcache should prevent needing to rewind in the stream), but it won't break anything so why not? :)

PR #904, opened 2025-01-07
Added option to differentiate between parse errors and black lines

mpownby:
For ld-process-vbi, I added an option to differentiate between parse errors and black lines.
Differentiating between parse errors and black lines is useful because if parse errors are known, they can be auto-fixed by studying the surrounding picture number cadence.

### DIFF
--- a/lddecode/core.py
+++ b/lddecode/core.py
@@ -39,8 +39,11 @@ except ImportError:
     # If not running Anaconda, we don't care that mkl doesn't exist.
     pass
 
-# XXX: This is a hack so that logging is treated the same way in both this
-# and ld-decode.  Probably should just bring all logging in here...
+# Uncomment (and add to) to get full traces of numpy warnings
+#import warnings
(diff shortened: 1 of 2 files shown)


ifb:
Is there a reason this should not be enabled all the time?  Why make it an option?

mpownby:
> Is there a reason this should not be enabled all the time? Why make it an option?

I was just trying to be conservative and not break anyone who may be relying on the old behavior (not that I can think of a reason why anyone would hehe).

ifb:
I'm not scared of breaking changes, but others can chime in on that.  Making this a simple 10-line patch beats making this new behavior optional, IMHO.

Affected user(s) can find/replace the -1 with 0.

harrypm:
I don't think it would break anything to make it a hard patch in the tool-chain, as long as it does not magically break VITC or CC reading its fine.

happycube:
Yeah, go ahead and enable this all the time.

mpownby:
> Yeah, go ahead and enable this all the time.

done.

PR #911, opened 2025-03-04
Analog Front End (AFE) Gnuradio Playground

VideoMem:
Added a ZMQ REQ / REP for back and forth communication from vhs-decode and gnuradio.

Using the --gnrc switch flag in vhs-decode, it sends the uncompressed 40MSPS RAW RF float data and receives it back prior to demodulation (FM decoding, TBC and chroma processing).

It is intended to use with gnuradio-companion as a playground to find the best RF EQ, with hopes of troubleshooting the decoding ringing.

The use is very simple.

Decode adding -t1 and --gnrc flag on you decode command line.

It will print a message like this:

```text
Initializing ZMQSend (REP) at pid 64926, port 5555
Initializing ZMQReceive (REQ) at pid 64926, port 5556
Open GNURadio with ZMQ REQ source set at tcp://localhost:5555 and ZMQ REP sink set at tcp://*:5556
The data stream will be of the float type at 40MSPS (40MHz sample rate)
It will send the raw RF for further processing prior to demodulation (useful for RF EQ discovery and group delay compensation)
You might want to do this in single threaded decode mode (-t 1 parameter)
```

Then open the graph located on tools/zmq_afe/ZMQ_AFE.grc

image

It comes with a builtin frequency spectrum
image

Whatever you transform the signal between the ZMQ REQ source and the ZMQ REP sink it will be decoded back by vhs-decode.
Currently no transforms are applied


VideoMem:
Sorry, wrong repo

PR #913, opened 2025-03-08
Efm decoder 20250308

simoninns:
This is the new suite of super-duper EFM tools :)  I have put the documentation into the wiki already (side bar->efm-decoder).  Note that this does not replace ld-decode-efm yet as it requires much more 'real world' testing with different EFM images.  In my testing it consistently performs better than ld-decode-efm though.

The tools also include an F2 Frame level stacker which can combine multiple EFM images resulting in greatly reduced errors.  Another big plus is the audio output produces both a WAV and an Audacity metadata label file that shows track boundaries, conceals and silences within the output audio (with CDDA timestamps)

Mostly, this is a ground up re-write to bring structure and sanity to the EFM decoder.  The code is slower than ld-decode-efm, but it is much better organised and easy to understand.  Also, all parts are divided up to allow improvements on individual parts without having to understand the whole process.

Testing on the data side of things (i.e. Domesday AIV) has shown that PAL EFM decoding by ld-decode is far, far worse than NTSC audio discs.  I'm not sure why this is, but these new tools provide detailed statistics and debug to assist in optimising the upstream ld-decode also.

The tools should be considered beta at best - but, since they install alongside the existing tools - I don't see any harm in including them.

One final note: the expwd CIRC library is included as a git sub-module for easier maintenance.  The old copy of the library is still there for ld-process-efm.  This is better in all respects except that, when cloning the overall repo, it is necessary to manually update the submodules to pull them in:  git submodule update --init

Any questions, just let me know!


happycube (approved):
i should do a full review, but i just wanna commit it.

atsampson (approved):

atsampson:
It's missing a #include for QLocale now.

simoninns:
I've done basically no real speed optimization yet - it probably could be a lot faster; but I was purely focused  on the structure and functionality.  ld-process-efm is a mess, this suite is (I hope!) much easier to understand and work with.  I think I've got all the Qt6 stuff nailed down now so, once the CI passes, I'll merge it.  After weeks of working on this I'm almost blind to it now :D so I want more hands and eyeballs on it.  Thanks for the reviews!

PR #915, opened 2025-03-16
Fix I being undefined and some typos

dp111:


simoninns (changes requested):
This .gitignore change looks like a mistake? Please revert the file and then it will be possible to merge.

PR #922, opened 2025-04-24
Y/C and mono decoding for ld-analyse + QOL improvements.

vrunk11:
This update reworks diferent par off ld-analyse : 
- the sizing of element of the main windows adapt dynamically.
- the window resize to fit perfectly then content.
- default parameter when loading a source adapt setting for optimal use depending on Y/C vs CVBS.
- luma nr set to 0 by default with checkbox for disabling and comparing quickly the NR
- ld-analyse now have the mono decoder.
- the mono decoder now support luma NR in ld-analyse and in the chroma-decoder
- Support for native Y/C preview and a checkbox for allowing combining or not Y/C to CVBS.

### DIFF
--- a/tools/ld-analyse/mainwindow.cpp
+++ b/tools/ld-analyse/mainwindow.cpp
@@ -187,7 +187,7 @@ void MainWindow::resetGui()
 
     ui->posNumberSpinBox->setValue(1);
     ui->posHorizontalSlider->setValue(1);
-    ui->dropoutsPushButton->setText(tr("Dropouts Off"));
+   (this->width() >= 930) ? ui->dropoutsPushButton->setText(tr("Dropouts Off")) : ui->dropoutsPushButton->setText(tr("Drop N"));
 
     setViewValues();
 
@@ -201,10 +201,15 @@ void MainWindow::resetGui()
 
     // Set option button states
     ui->videoPushButton->setText(tr("Source"));
-    displayAspectRatio = false;
+    displayAspectRatio = true;
     updateAspectPushButton();
     updateSourcesPushButton();
-    ui->fieldOrderPushButton->setText(tr("Normal Field-order"));
+    if (this->width() > 1000)
+		ui->fieldOrderPushButton->setText(tr("Normal Field-order"));
+	else if (this->width() >= 930)
+		ui->fieldOrderPushButton->setText(tr("Normal order"));
+	else
+		ui->fieldOrderPushButton->setText(tr("Normal"));
 
     // Zoom button options
     ui->zoomInPushButton->setAutoRepeat(true);
@@ -264,6 +269,9 @@ void MainWindow::updateGuiLoaded()
 
     // Disable "Save JSON", now we've loaded the metadata into the GUI
     ui->actionSave_JSON->setEnabled(false);
+	
+	//resize the windows to fit the content in full screen
+	MainWindow::resize_on_aspect();
 }
 
 // Method to update the GUI when a file is unloaded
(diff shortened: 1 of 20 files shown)


vrunk11:
image
image

eshaz:
Maybe an alternative to changing the global defaults would be to write a feature that allows users to save defaults for the ui layout to their machine.

harrypm:
Practically everyone changes the 1:1 to 4:3 so it makes sense to keep it as 4:3 default because that's what the majority of users and even myself use.

It definitely would be nice to have a config file system for preferences.

oyvindln:
>Practically everyone changes the 1:1 to 4:3 so it makes sense to keep it as 4:3 default because that's what the majority of users and even myself use.

Well I don't, but agree it should ideally save preferences so everyone can keep their prefereces

vrunk11:
> Y/C mode works really well.
> 
> A few things that I noticed that need to be fixed.
> 
>     * The window size changes each time ld-analyse is opened, and shrinks below the 200% zoom that I use. This should be fixed so that the window only resizes down the fit the visible video. Or maybe the window should just remain at the size the user last used like it does in the current build.
> 
>     * "Phase compensating decoder" should default to checked, looks like this broke somewhere since it's no longer defaulted. This is needed for VHS and tape chroma decoding.
> 
>     * Looks like the pixel ratio mode has changed from SAR 1:1 to DAR 4:3. It might be a little easier to see the video and make adjustments with square pixels, so maybe that should remain defaulted to 1:1.

For the zoom i efectively tought it was not practical to resize to those size as it seem way to big for even an 1440p screen. but efectively i could lock resize over a ceetain size or take time to compute the good resize for the different zoom but that would be tedious.

for the phase compensation its effectively did an update cause i found it was non consistent between cvbs and y/c and the phase compensation seem to give color that was more wrong so i decided it was probably better to set it to false by default everywhere to avoid user say the color look different or wrong. but i never dealed with NTSC vhs. maybe there is something in the metadata i could use to know if its a tape and not just a high quality Y/C source ?

vrunk11:
the setting saving could be a good idea but would have to be on another update. and i feel the default should already fit the most typical case.

tools/ld-analyse/mainwindow.cpp
```diff
@@ -994,21 +1067,75 @@ void MainWindow::on_aspectPushButton_clicked()
...
+			else if (displayAspectRatio && tbcSource.getIsWidescreen())// 16:9
+			{
+				this->resize(1052, 665);
+			}
+			else// 1:1
+			{
+				this->resize(930, 665);
+			}
+		}
+	}
```
eshaz, line 1111: This seems to be what is running that unexpectedly resizes the window when I open ld-analyse. This is pretty jarring, and I need to resize the window back to where I had it each time. I would suggest we fix this, or remove it and add it in later when it's working. It'd be nice to see the chroma decoder changes go in, so I would think the latter option is preferred.

Also, where do these pixel values come from? However they are derived, the window size should be calculated relative to size of the video. The only constant here is the actual size of the video. The UI elements around the window can be different sizes depending on themes, OS, zoom values, etc.

vrunk11: > This seems to be what is running that unexpectedly resizes the window when I open ld-analyse. This is pretty jarring, and I need to resize the window back to where I had it each time. I would suggest we fix this, or remove it and add it in later when it's working. It'd be nice to see the chroma decoder changes go in, so I would think the latter option is preferred.
> 
> Also, where do these pixel values come from? However they are derived, the window size should be calculated relative to size of the video. The only constant here is the actual size of the video. The UI elements around the window can be different sizes depending on themes, OS, zoom values, etc.

Yes i wrote that code. and you issue because i never dealed with zoom. the size should be the same on all OS as its handled by qt. the only things that might potentialy change is the design off the button but that shouldnt impact any sizing. for finding the size i just did trial and error until it fit perfectly. i could look into finding the perfect size for the x2 zoom. and for x3 maybe i would see 🤔. but for non integer size its wouldnt be possible.

eshaz: If there's a way to get the pixel size height and width of the inner video image, that could be used to resize the window. Basically instead of hardcoding the pixel values here, they would be derived from the size of the video image and the border between the view area and the main window.

Pseudo code would be something like this:
```
Get the height and width of the scaled video image
Get the height and width of the video view area
Get the height and width of the main window

// account for the size of the ui elements
Outer height and width = size of the main window - size of the view area

// calculate window size relative to video image and window ui elements
New size = video image size + outer size

Resize the window to the new size
```

vrunk11: i thinks they both scale differently but im going to do few tests and see

vrunk11: i fixed it it was a lot easier than i thought the margin is always the same. now it resize perfectly for all zoom size.

vrunk11:
image
this should fix the phase compensation issue mentioned before. It now enable it for NTSC tape by default.

tools/ld-analyse/mainwindow.cpp
eshaz: Confirmed, works much better now with zoom.

PR #923, opened 2025-05-01
Fix simple typos found with codespell

dp111:


tools/ld-disc-stacker/stacker.cpp
```diff
@@ -768,7 +768,7 @@ inline bool Stacker::isDropout(const DropOuts& dropOuts, const qint32 fieldX, co
     return false;
 }
 
-// Method returns true if specified pixel is a dropout
+// Method returns true all specified pixel are dropouts
```
atsampson, line 771: That one should be "if all"...

atsampson:
Changes to the C++ code all look good, other than the one I've added a note on above. It's mostly removing spaces at the end of lines.

happycube (approved):

Issue #918, opened 2025-03-18
efm-decoder-f2: SectionTime invalid frame value crash


JoeyLemur:
### Description

This disc (disc 1 of https://www.lddb.com/laserdisc/24880/G128F0083/Last-Starfighter-The) fails with "Fatal: SectionTime::SectionTime(): Invalid frame value of -269999" when run against efm-decoder-f2

### Steps to Reproduce

1. `ld-decode --start 350 --length 1800 --threads 6 --ntsc the_last_starfighter_s1.flac.ldf test/output`
2. `cd test`
3. `efm-decoder-f2 --debug --show-all-debug output.efm output.f2`

### Actual Behaviour

Gets as far as "F2 Section Metadata Correction statistics" before crashing:
```
Info: F2 Section Metadata Correction statistics:
Info:   F2 Sections:
Info:     Total: 0 (0 F2)
Info:     Corrected: 0
Info:     Uncorrectable: 0
Info:     Pre-Leadin: 4504
Info:     Missing: 0
Info:     Padding: 0
Info:     Out of order: 0
Info:   QMode Sections:
Info:     QMode 1 (CD Data): 0
Info:     QMode 2 (Catalogue No.): 0
Info:     QMode 3 (ISO 3901 ISRC): 0
Info:     QMode 4 (LD Data): 0
Info:   Absolute Time:
Info:     Start time: 59:59:74
Info:     End time: 00:00:00
Fatal: SectionTime::SectionTime(): Invalid frame value of -269999
```

No output is written to `output.f2`

### Environment
- `efm-decoder-f2 Branch: main / Commit: 5d339d9d`
- MacOS 15.3.2 (MacBook Air M2, 2022)

JoeyLemur:
[output.efm.gz](https://github.com/user-attachments/files/19321714/output.efm.gz)

[debug.log](https://github.com/user-attachments/files/19321716/debug.log)

JoeyLemur:
Neglected to mention the actual crash error:  `Abort trap: 6`

JoeyLemur:
Different disc, same issue.

[test2_debug.log](https://github.com/user-attachments/files/19326258/test2_debug.log)
[test2.efm.gz](https://github.com/user-attachments/files/19326257/test2.efm.gz)

boysrbackintown:
[...]

PR #941, opened 2025-11-05
Fix for issue #918 - add support for ECMA-130 non-compliant EFM data that lacks timecodes

simoninns:
A fix for issue #918 which firstly detects when the decoding fails and outputs a helpful prompt (rather than a core-dump).

The efm-decoder-f2 now has a --no-timecode option which ignores the missing timecode and instead populates the sector metadata with the expected timecode (generating a single track that contains all EFM contents).  This, of course, means that time-code correction cannot be used and may therefore result in data loss when used on EFM data that contains timecodes (why you would do that though, is anyone's guess...)

If this works generally then I recommend removing the old efm-decoder completely as the new decoder should have functional equivalence now.

I've also fixed some minor issues with debug output across the EFM decoder tool-chain but I'm not dead-sure this is local to Fedora 43 environments - the changes shouldn't cause any Ubuntu issues though.

### DIFF
--- a/tools/efm-decoder/CMakeLists.txt
+++ b/tools/efm-decoder/CMakeLists.txt
@@ -3,10 +3,6 @@ cmake_minimum_required(VERSION 3.10)
 # Set the project name
 project(efm-tools)
 
-# Specify the C++ standard
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED True)
-
 # Handle Qt version selection - default to Qt5 for backward compatibility
 if(NOT DEFINED USE_QT_VERSION)
     set(USE_QT_VERSION 5)
--- a/tools/efm-decoder/tools/efm-decoder-f2/src/decoders/dec_f2sectioncorrection.h
+++ b/tools/efm-decoder/tools/efm-decoder-f2/src/decoders/dec_f2sectioncorrection.h
@@ -64,8 +64,6 @@ private:
     qint32 m_maximumGapSize;
     qint32 m_paddingWatermark;
 
-    bool m_noTimecodes;
-
     // Statistics
     quint32 m_totalSections;
     quint32 m_correctedSections;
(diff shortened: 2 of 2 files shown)


tools/efm-decoder/tools/efm-decoder-f2/src/decoders/dec_f2sectioncorrection.cpp
```diff
@@ -40,7 +40,8 @@ F2SectionCorrection::F2SectionCorrection()
       m_qmode3Sections(0),
       m_qmode4Sections(0),
       m_absoluteStartTime(59, 59, 74),
-      m_absoluteEndTime(0, 0, 0)
+      m_absoluteEndTime(0, 0, 0),
+      m_noTimecodes(false)
```
atsampson: This produces a warning from GCC 15, because the order of initialisation here doesn't match the order in which these member variables are declared in the class:

```
In file included from /home/ats/src/ld-decode/tools/efm-decoder/tools/efm-decoder-f2/src/decoders/dec_f2sectioncorrection.cpp:25:
/home/ats/src/ld-decode/tools/efm-decoder/tools/efm-decoder-f2/src/decoders/dec_f2sectioncorrection.h: In constructor ‘F2SectionCorrection::F2SectionCorrection()’:
/home/ats/src/ld-decode/tools/efm-decoder/tools/efm-decoder-f2/src/decoders/dec_f2sectioncorrection.h:85:17: warning: ‘F2SectionCorrection::m_absoluteEndTime’ will be initialized after [-Wreorder]
   85 |     SectionTime m_absoluteEndTime;
      |                 ^~~~~~~~~~~~~~~~~
/home/ats/src/ld-decode/tools/efm-decoder/tools/efm-decoder-f2/src/decoders/dec_f2sectioncorrection.h:67:10: warning:   ‘bool F2SectionCorrection::m_noTimecode ’ [-Wreorder]
   67 |     bool m_noTimecodes;
      |          ^~~~~~~~~~~~~
/home/ats/src/ld-decode/tools/efm-decoder/tools/efm-decoder-f2/src/decoders/dec_f2sectioncorrection.cpp:27:1: warning:   when initialized here [-Wreorder]
   27 | F2SectionCorrection::F2SectionCorrection()
      | ^~~~~~~~~~~~~~~~~~~
```

Moving the declaration of `m_noTimecodes` to below the times would fix it.

simoninns: Fixed in additional commit to PR

Issue #950, opened 2025-11-22
PAL CAV discs failing to decode


simoninns:
### Bug Description

PAL CAV Pioneer discs do not decode. Tested with the following discs:

- Asas de Portugal - Rolling in the sky
- Blue Angels - Rolling in the sky
- Blue Impulse - Rolling in the sky
- Patrouille de France - Rolling in the sky
- Ram-K Su-27
- Red Arrows - Rolling in the sky
- Snowbirds - Rolling in the sky

### Steps to Reproduce

I have prepared a short sample from the Red Arrow disc around frames 14950-15050 to demonstrate the issue.

Project SFTP: simoninns/PAL_issue/Red_Arrows_side1_2025-11-22_08-35-33.ldf

### Expected Behaviour

Disc decodes - tbc is produced

### Actual Behaviour

<img width="1160" height="670" alt="Image" src="https://github.com/user-attachments/assets/288939df-66fc-4c24-9761-57d7208ec6bb" />

### Environment

- Decode version: 88f9c53da99b26759afcc89f585d833309ed0218
- Operating System: Fedora 43 - Python 3.12 venv
- DdD Latest firmware and Linux application from github

Player RF output measured at test point from Red Arrows frame 15000 using PicoScope 100MHz DSO (approx 320mV PTP):

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/b8308f14-e491-4c83-ae50-e287cf4361e9" />

Captured RF unpacked into 16-bit RIFF and shown in Audacity:

<img width="2901" height="1445" alt="Image" src="https://github.com/user-attachments/assets/65d4154b-cd7d-482a-ae68-8ef6740817c7" />

### Additional Information

Player and capture process was tested with GGV1069 prior to capture. GGV1069 capture and decode was successful SNR levels as expected from calibrated capture LDV4300D (Domesday Duplicator 1). I also tested with a PAL CLV disc (Grosse Pointe Blank) - both sides captured and decoded correctly.

Issue #949, opened 2025-11-21
Regression: Worse Brightness differences between the fields than before


jsaowji:
### Checklist

- [x] I have searched the issues page for any duplicate issues open or closed and confirmed that this bug has not been reported before.
- [x] I have tested the issue with the current build.
- [x] I have attached log files, uploaded sample data, and commands used so that the issue can be easily reproduced by the developers.

### Bug Description
For some NTSC laserdisc the the AGC doesn't work as good as before https://github.com/happycube/ld-decode/pull/934 was merged. There are brightness differences between the fields that weren't there before 

It doesn't seem that bad on most of my discs but for the provided sample its very noticable. 

### Steps to Reproduce

1. git checkout fc82480b92e7f8c4ddf405f63c7db84a399ec748 (or master); build software; ld-decode o.ldf bad
2. git checkout 116c00c0b9204d73d6cb8586056b2153dfa7904c; build software; ld-decode o.ldf good
3. Look at both files with ld-analyse

### Expected Behaviour

Frame looks white

### Actual Behaviour
The individual fields have different brightness

### Environment

- Decode version: view Reproduce
- Operating System: Linux 
- Hardware Used: not relevant

### Additional Information

<img width="414" height="341" alt="Image" src="https://github.com/user-attachments/assets/ab39a576-3397-4654-876f-88ddc29ea71f" />
<img width="458" height="452" alt="Image" src="https://github.com/user-attachments/assets/90677b58-8ebf-4ad4-8e85-06dfbef18de8" />

[o.ldf.zip](https://github.com/user-attachments/files/23671288/o.ldf.zip)

### Is it related to tbc-video-export?
no

happycube:
I'm reverting the change for now since I'm not sure when I can debug it.  I'll merge a pull request back in if [eshaz](https://github.com/eshaz) fixes it first.

PR #952, opened 2025-11-22
Revert "use interpolation to calculate wow, scale using interpolation function"

happycube:
Reverts happycube/ld-decode#934, causes issue #950. (edit:  I meant 949.)


simoninns:
are you sure you mean issue #950 ?

happycube:
Oops, I meant #949

Issue #946, opened 2025-11-08
Y/C Scanline Ossiliscope Support


harrypm:
The current ld-analyse code has an major issue of not loading both Y & C data properly when the scan-line oscilloscope is being enabled on duel/s-video style tbc sets.

This was noted by Brad on the DD86 discord [Message Chain Link](https://discord.com/channels/665557267189334046/665834485975351307/1433627588969889973)

Date: 31 October 2025

----------------------

Y/C Source selection. YC, Y, C tickboxes

<img width="1172" height="946" alt="Image" src="https://github.com/user-attachments/assets/6b7a4191-7a1d-4d19-a350-ad8a08c70532" />

<img width="1172" height="946" alt="Image" src="https://github.com/user-attachments/assets/adaa6c88-eccc-4b6d-b40e-9cf37673c75d" />

<img width="1172" height="946" alt="Image" src="https://github.com/user-attachments/assets/2bdebd33-d7b4-4375-9529-861db91d5f65" />

----------------------

Y Source selection. YC, Y, C tickboxes

<img width="1172" height="946" alt="Image" src="https://github.com/user-attachments/assets/e5ae985c-10e4-44b7-ad02-9a90ec76c008" />

<img width="1172" height="946" alt="Image" src="https://github.com/user-attachments/assets/6f925417-a5f1-4355-9cb1-116578585f28" />

<img width="1172" height="946" alt="Image" src="https://github.com/user-attachments/assets/400b1b78-2dd6-42ae-ace7-976360aa917a" />

----------------------

C Source selection. YC, Y, C tickboxes

<img width="586" height="473" alt="Image" src="https://github.com/user-attachments/assets/15bbc3ad-98d4-49d2-a3e2-80389ddc2451" />

<img width="586" height="473" alt="Image" src="https://github.com/user-attachments/assets/731f23f9-b1c0-4545-b9e3-92ce4f4d0653" />

<img width="586" height="473" alt="Image" src="https://github.com/user-attachments/assets/566470b3-ee0e-4fb2-bd3f-012897d41095" />

----------------------

What I really don't understand here is that the "Y" and "C" tickbox oscilloscopes on the Y Source selection differ very slightly from the Y/C Sources selection. 

The "YC" oscilloscope is identical for both views, as expected..

I see what it is now. Setting the chroma decoder to Mono causes "Y" and "C" to also be identical between Y Source and Y/C Sources.

[...]

PR #957, opened 2025-11-25
Fix for scope issue on Y/C input

oyvindln:
Fixing issue #946


harrypm:
Still not loading properly with Y/C Enabled

Screenshot from 2025-11-28 02-49-16

Screenshot from 2025-11-28 02-49-21

Screenshot from 2025-11-28 02-49-30

Screenshot from 2025-11-28 02-49-39


Screenshot from 2025-11-28 02-49-45

harrypm:
That's still incorrect mapping compared to how CVBS is handled.

Y/C is still not showing both channels worth of information.

I.g 



My current test branch produces an accurate Y/C mode output with S-Video style TBCs in Y+C / Chroma mode.

https://github.com/harrypm/ld-decode/tree/ld-analyse/improvements

I'll update this and add an example, but the QT scaling is still broken on Windows builds.

Issue #951, opened 2025-11-22
Modify ld-decode tools to use SQLite rather than JSON


simoninns:
Currently ld-decode produces a JSON metadata file whilst decoding LaserDisc samples.  Each tool reads the entire JSON file and adds additional metadata as the processing continues.

This proposal is only for the Qt/C++ tools.  Therefore these are the proposed layers:

- Upstream producer ld-decode (unchanged)
  - Still emits JSON metadata
- Ingest/converter app (new)
  - Reads incoming JSON (from .json file)
  - Validates it according to current documentation
  - Writes to SQLite .db file using a well-defined schema
- ld-decode tools shared library is modified
  - Used by all downstream tools for metadata handling
  - Stops reading/writing JSON files and instead talks only to SQLite
  - Keeps the same public C++ API so downstream tools don’t know or care that storage changed

Of course, ld-decode can at any time move to SQLite also (and the converter can be removed) - but, for now, decoupling and reducing complexity is desirable.

The SQL schema should be strongly typed.  This will ensure a clear structure and improve the long-term maintainability of the code.  The converter tool will act as the gatekeeper for this and reject any further changes to the ingested JSON metadata.

Note that concurrent access to the SQL database should be considered.  However, it is not currently required as there is no support for this in the current JSON based architecture. Concurrency is interesting though as it would allow tools like ld-analyse to live-update as other tools process the TBC and associated items.

This is quite a bit of work (and will eventually effect ld-decode itself) - so I welcome comments and feed back on this proposal before I start working on it.

simoninns:
This is the compete updated schema:


```
------------------------------------------------------------------
-- Schema Versioning
------------------------------------------------------------------
PRAGMA user_version = 1;

------------------------------------------------------------------
-- 1. Capture-level metadata (one per JSON file)
------------------------------------------------------------------
CREATE TABLE capture (
    capture_id INTEGER PRIMARY KEY,
    system TEXT NOT NULL
        CHECK (system IN ('NTSC','PAL','PAL_M')),
    video_sample_rate REAL,
    active_video_start INTEGER,
    active_video_end INTEGER,
    field_width INTEGER,
    field_height INTEGER,
    colour_burst_start INTEGER,
    colour_burst_end INTEGER,
    is_mapped INTEGER
        CHECK (is_mapped IN (0,1)),
    is_subcarrier_locked INTEGER
        CHECK (is_subcarrier_locked IN (0,1)),
    is_widescreen INTEGER
[...]

PR #960, opened 2025-11-29
Migrate metadata storage from JSON to SQLite

simoninns:
- Added ld-json-converter tool for migrating existing JSON metadata to SQLite format .db files
- Implemented SQLite I/O library with optimized bulk read operations
- Updated all tools to use SQLite metadata format (.db files) instead of JSON
- Modified ld-analyse, ld-chroma-decoder, ld-disc-stacker, ld-discmap, ld-dropout-correct, ld-export-metadata, ld-process-vbi, and ld-process-vits
- Fixed thread-safe database connections for ld-analyse
- Removed Wayland support warnings (from ld-analyse)
- Enhanced error handling and performance optimizations

This migration provides better performance, reduced memory usage, and improved reliability for metadata operations via SQLite.

This PR solves issue #951 

The schema and translation information (from JSON) is included in the README.md within the ld-json-converter tool directory.

NOTE: This PR will make all tools expect .db files in SQLite database format.  Any ld-decode'd TBC files will need to have the JSON processed by ld-json-converter before any tools will work.  This PR is the ground work for changing ld-decode to SQLite also (removing the need for the intermediate step).  Obviously this will be a little annoying to users that like to ride the bleeding-edge of the github - but that's life).


simoninns:
Looks like the CI/CD stuff is still expecting JSON files - I will look into how to fix that and reissue the PR when fixed

PR #962, opened 2025-11-30
Migrate metadata storage from JSON to SQLite

simoninns:
This PR makes a fundamental change to the metadata handling in ld-decode moving from JSON to SQLite (3).

This PR fixes issue #951 

Since ld-decode.py doesn't yet support SQLite - there is a ld-json-converter tool that converts the output from ld-decode.py into SQLite for the tool-chain (i.e. takes .tbc.json and outputs .tbc.db).

Another important change is that field numbers in the metadata are now zero-indexed.  In all cases the TBC library is maintained as much as possible to ensure minimal changes to the tools themselves.

Obviously changing the metadata format has a large impact on the test-suite also - so this has been updated and maintains the same testing but also uses ld-json-converter to convert the test-data metadata into the new format.

Clearly this change touches many points in the tool chain - I've tested as much as I could, but there might be edge-cases that I and the CI didn't find.  Since the first step to using the new tools is to convert the original JSON, the risk is low as the original JSON can always be converted again.

Here are some of the highlights:

- Added ld-json-converter tool for migrating existing JSON metadata to SQLite format .db files
- Implemented SQLite I/O library with optimized bulk read operations
- Updated all tools to use SQLite metadata format (.db files) instead of JSON
- Modified ld-analyse, ld-chroma-decoder, ld-disc-stacker, ld-discmap, ld-dropout-correct, ld-export-metadata, ld-process-vbi, and ld-process-vits
- Fixed thread-safe database connections for ld-analyse
- Removed Wayland support warnings
- Enhanced error handling and performance optimizations

Added ld-json-converter to the test-script to get the CI/CD running
Change ld-json-converter prompt
Fix CI: Add SQLite dependencies and correct ld-json-converter arguments

- Add libqt5sql5-sqlite for Qt5 build
- Add libqt6sql6-sqlite for Qt6 build
- Add libsqlite3-dev for both builds
- Fix ld-json-converter to use --input-json and --output-sqlite arguments

Fixed bug in ld-json-convert that caused the CI tests to fail
Fixing some bugs in both the testing and the tools - checking in to make sure the changes don't get too confusing
Fixed bug in the SQLite library
Expected VBI reversed by mistake. Now corrected
Fixed off by one error in encoders (caused by change in metadata indexing) and fixed test issues

### DIFF
--- a/tools/ld-json-converter/jsonconverter.cpp
+++ b/tools/ld-json-converter/jsonconverter.cpp
@@ -656,9 +656,21 @@ bool JsonConverter::insertData(LdDecodeMetaData &metaData)
                 query.bindValue(3, dropouts.endx(i));
                 
                 if (!query.exec()) {
-                    qCritical() << "Failed to insert dropout data for field" << fieldNum << "dropout" << i << ":" << query.lastError().text();
-                    m_database.rollback();
-                    return false;
+                    // Check if this is a UNIQUE constraint violation (duplicate dropout)
+                    QString errorText = query.lastError().text();
+                    if (errorText.contains("UNIQUE constraint failed", Qt::CaseInsensitive)) {
+                        qInfo() << "Skipping duplicate dropout in field" << fieldNum << "dropout" << i << ": "
+                                << "fieldLine=" << dropouts.fieldLine(i) 
+                                << "startx=" << dropouts.startx(i) 
+                                << "endx=" << dropouts.endx(i);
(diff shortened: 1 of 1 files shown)


simoninns:
I've merged the changes to main into the PR - however, there seems to be some odd behaviour around the JSON converter and some tape sources which I need to look into; so I will look into that before merging this PR

PR #973, opened 2025-12-06
Added and amended README.md files for all tools 

simoninns:
The documentation is sliding out of date from the tools because the wiki and the source code are decoupled.  So I've made README.md files for all the tools (and parts of the tools directory structure) and harmonized the contents.

I've used some AI magic to ensure that the source and readme documentation matches - but there might be some funnies in there - I did a fast manual check, but it's more important to have the documentation in place and ready for update than missing completely.

ofc there is the question of harmonizing this with the wiki... but that's a separate issue.


tools/ld-process-efm/README.md
```diff
@@ -0,0 +1,258 @@
...
+- Use `--sample-rate` to override if needed
+- Check disc specifications
+
+## LaserDisc Digital Audio
+
+### Format Types
+- **PCM Audio**: Highest quality, occupies video lines
+- **AC-3 (Dolby Digital)**: Compressed surround (use ld-process-ac3)
+- **Analog Audio**: Standard LaserDisc audio (decoded separately)
+
```
harrypm: DTS is missing

simoninns: This whole readme was a mistake... I've just changed it to say "use efm-decoder-*" instead... no point in documenting something that I'm going to remove soon :)

PR #1032, opened 2026-04-06
Upstream changes

oyvindln:
Merge in all upstream changes - this should be ready to go now

once this is done we need some changes in ld-decode so we don't need the janky workaround that deletes the sqlite file after creation when not using sqlite mode


oyvindln:
wait I meant to merge this into vhs-decode not here

PR #1034, opened 2026-04-09
Add AC3 RF audio decoding via ac3rf Python extension

staffanu:
LaserDisc discs containing Dolby AC3 audio encode the audio as a QPSK-modulated
signal at 2.88 MHz on the video carrier. This PR adds support for demodulating
that signal and writing the raw QPSK symbols to an `.ac3sym` file, analogous to
how EFM T-values are written to `.efm`.

- Removes the old dead AC3 pipe code (`ac3_pipe`, `AC3filter`) which referenced
  `ld-ac3-demodulate`/`ld-ac3-decode` executables that no longer exist
- Replaces it with `AC3demodulate()` using the `ac3rf` Python extension
  (built from [ldaudio/ac3rf-decode](https://bitbucket.org/staffanulfberg/ldaudio),
  publicly readable over HTTPS)
- When `--ac3` is passed, ld-decode writes a `.ac3sym` file containing raw QPSK
  dibits (1 byte/symbol) and records the per-field symbol count in the `.tbc` DB
  as `ac3_symbols`
- The Nix flake builds the `ac3rf` Python extension from a pinned commit of the
  ldaudio repo (which pulls in the `firpm` submodule via `?submodules=1`);
  `nanobind` is pre-fetched as a flake input so the build requires no network
  access inside the Nix sandbox

The `.ac3sym` output is intended to be consumed by a downstream decoder to
produce a final `.ac3` bitstream. The per-field symbol counts stored in the DB
allow the downstream tool to reconstruct per-field byte offsets into the file.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -18,6 +18,43 @@
         "type": "github"
       }
     },
+    "ldaudio": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1775267416,
+        "narHash": "sha256-moCkqtTp1D2crJpHFX6ND8UTbpF7zgBKanZPxLLYoAU=",
+        "ref": "refs/heads/master",
+        "rev": "1d976d7e1f095c4c24d02c3dfa3804d4208da628",
+        "revCount": 465,
+        "submodules": true,
+        "type": "git",
+        "url": "https://bitbucket.org/staffanulfberg/ldaudio"
+      },
+      "original": {
+        "submodules": true,
+        "type": "git",
+        "url": "https://bitbucket.org/staffanulfberg/ldaudio"
+      }
+    },
+    "nanobind-src": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1733439979,
+        "narHash": "sha256-9OpDsjFEeJGtbti4Q9HHl78XaGf8M3lG4ukvHCMzyMU=",
+        "ref": "refs/tags/v2.4.0",
+        "rev": "0f9ce749b257fdfe701edb3cf6f7027ba029434a",
+        "revCount": 1015,
+        "submodules": true,
+        "type": "git",
+        "url": "https://github.com/wjakob/nanobind"
+      },
+      "original": {
+        "ref": "refs/tags/v2.4.0",
+        "submodules": true,
+        "type": "git",
+        "url": "https://github.com/wjakob/nanobind"
+      }
+    },
     "nixpkgs": {
       "locked": {
         "lastModified": 1774799055,
@@ -37,6 +74,8 @@
     "root": {
       "inputs": {
         "flake-utils": "flake-utils",
+        "ldaudio": "ldaudio",
+        "nanobind-src": "nanobind-src",
         "nixpkgs": "nixpkgs"
       }
     },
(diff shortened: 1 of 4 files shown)


happycube:
Sorry I didn't notice this for so long!

staffanu:
No problem! Glad you merged it and hope it works!

simoninns:
This PR caused CI/CD failure and had to be reverted - it will need to be resubmitted against the current commit for inclusion (I put the cause in the notes of the PR to revert).

On a related note, this PR was pulling in a nix build from a remote repo (a nix submodule) which had a different nix package base (unstable rather than 25.11) - The architecture of this would be a lot better if it followed the EFM implementation and kept the demod within ld-decode's python base rather than coupling it externally.

@happycube 's call really - but I don't think the AC3 demod in a remote flake is a strong approach imho

staffanu:
Sorry for the nix version problem.  I can easily fix this and submit a new PR.  BTW, I started out a new repo for the ac3rf/efm demodulation and muse player which is significantly smaller than the old repo, and lots of old scala code and hardware (kicad and vhdl) is gone, so a new PR will take in a smaller dependency.

That said, I'd be happy to instead copy the necessary files from a specific verision over instead. What do you mean by "keeping the demod within ld-decode's python base"?  Are you suggesting to port to python entirely, or just to move the code there?  I think porting to python might turn out hard to maintain since I do not do original development in python. If you have a prefrence to use something else than nanobind, however, that's fine for me.

simoninns:
my thought was that it would be better for the AC3 demod to follow the same implementation pattern as the existing EFM demod within ld-decode.  Implementing it in another language and pulling it in as a nix package (basically the nix version of git submoduling) is overly complex and adds maintenance/testing complexity.  I understand that you might not want (or have the python knowledge) to implement it that way... but then it might be better to find someone who can port your code.

I'm just a little worried that spreading ld-decode's 'core' over multiple repos which different people own is not a good approach for the project.  @happycube is the owner of ld-decode though - so, if he's ok, that's all you need... but it might be worth having the discussion (at least on discord) before proceeding - rather than having PRs shelved and not merged.

staffanu:
I understand the concern, and as I said I'm happy to change the approach.  I'm not sure where on Discord this discussion first, but I'll jump in if I get a pointer! :)

So, the way I see it there are three ways to do this:

1. Use a flake to pull in code from my repo.  It definitely makes ld-decode less "standalone" and I understand that that might not be what everyone prefers. I did ask happycube about it and he said it was alright, but I'm not holding anyone to what they said if there are arguments for reconsidering.  For me this is obviously the easiest path.
2. "vendor" the required C++ code directly and build a python wrapper.  Makes the repo self contained, but still mixed language.  I haven't even looked around much but thought there was C++ code already in ld-decode.
3. Port the AC3 demodulation to python.  Nothing of the actual code is "hard" -- it is mostly FIR filtering and a DPLL. I can cheat my way in Python even if it isn't my preferred language, and with some help from AI I'm sure I could do it relatively quickly. I'm mostly concerned about the fact that any updates/improvements in my code base probably won't be propagated, at least not quickly. Of course, someone else could do it, but i have a feeling that might not happen any day soon.

PR #1050, opened 2026-07-14
Add AC3 RF audio demodulation in Python (lddecode/ac3rf.py)

staffanu:
## Summary

This is the resubmission of AC3 RF audio decoding (previously PR #1034, merged and then reverted in #1040). Following the discussion on #1034, the demodulator is now implemented in-tree in Python, following the same pattern as `lddecode/efm_pll.py` — numpy/scipy for filter design, numba for the DPLL and FIR kernels. There is no external repository, no nix flake dependency, and no new Python dependency (numpy/scipy/numba are already required).

With `--AC3` (NTSC only), ld-decode demodulates the 2.88 MHz QPSK subcarrier and writes the raw symbols to `<output>.ac3sym` (one symbol per byte, values 0–3). The number of symbols demodulated during each field is recorded in the field metadata as `ac3Symbols` / `ac3_symbols`, exactly analogous to `efmTValues` / `efm_t_values`. Framing, Reed-Solomon error correction and AC3 frame assembly happen downstream in decode-orc's AC3 RF Sink stage, which already queries this metadata column; a companion decode-orc PR restores that stage's TBC-source read path.

## What's in the PR

- `lddecode/ac3rf.py` — the demodulator: half-band Kaiser decimation stages, 2.88 MHz mixer, root-raised-cosine matched filter, and a DPLL for symbol timing recovery (numba jitclass, like `EFM_PLL`). Accepts arbitrary block lengths; all filter/timing state carries across calls.
- Integration in `core.py` replacing the dead `ac3_pipe` code (which shelled out to C++ tools removed in the removetools-202601 cleanup).
- `tests/test_ac3rf.py` — self-contained tests, no capture files needed: DQPSK loopback (clean + noisy), and an end-to-end test that synthesizes a complete NTSC LD signal (FM video carrier + AC3 subcarrier), runs `ld-decode --AC3` on it, and validates the demodulated symbols and the per-field metadata counts.
- Documentation of the `--AC3` behavior and the decode-orc handoff in `docs/user-guide/command.md`.

## Validation

- **Bit-exact vs the previous C++ implementation**: 100% symbol agreement over three real AC3 disc captures, at 40 MHz (int16) and 62.5 MHz (uint8) sample rates.
- **End-to-end on real discs**: `.ac3sym` output decoded through the AC3 framing/RS stage produces a clean AC3 elementary stream (48 kHz, 5.1, 384 kb/s), with near-zero Reed-Solomon corrections (e.g. 3 corrected symbols in 11,952 C1 codewords), and audio verified by listening.
- **Performance**: ~1.6× realtime single-threaded for 40 MSps input; runs `nogil` so it overlaps the decoder's other threads.

## Notes

- Compared to #1034, the metadata records per-field symbol *counts* rather than offsets, matching the EFM convention and what decode-orc's metadata readers query.
- The branch currently includes the small numpy 2.0 loader fix (PR #1049) because the end-to-end test synthesizes a `.s16` input; I'll rebase it out once that PR merges.

### DIFF
--- a/docs/user-guide/command.md
+++ b/docs/user-guide/command.md
@@ -368,8 +368,9 @@ Enable AC3-RF audio demodulation (NTSC only).  On AC3 LaserDiscs the
 analog right audio channel carries a QPSK signal at 2.88 MHz with Dolby
 Digital data at 288 kbaud.  With this option, ld-decode demodulates that
 signal (see `lddecode/ac3rf.py`) and writes the raw QPSK symbols to
-`output.ac3sym` (one symbol per byte, values 0-3); the symbol offset of
-each field is recorded in the field metadata.
+`output.ac3sym` (one symbol per byte, values 0-3); the number of symbols
+demodulated during each field is recorded in the field metadata
+(`ac3Symbols`, analogous to `efmTValues`).
 
 The `.ac3sym` file is not playable audio by itself: framing, Reed-Solomon
 error correction and AC3 frame assembly are performed downstream by
--- a/docs/user-guide/command.md
+++ b/docs/user-guide/command.md
@@ -364,7 +364,19 @@ ld-decode --preEFM input.ldf output
 ```
 
 #### `--AC3`
-Enable AC3 audio decoding (NTSC only).
+Enable AC3-RF audio demodulation (NTSC only).  On AC3 LaserDiscs the
+analog right audio channel carries a QPSK signal at 2.88 MHz with Dolby
+Digital data at 288 kbaud.  With this option, ld-decode demodulates that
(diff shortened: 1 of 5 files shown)


staffanu:
Pushed ef928ce8, syncing the DPLL loop gains with the reference C++ implementation (museld b82e92b): the proportional/integral gains are now computed from the second-order loop design parameters (1800 Hz natural frequency, damping 0.6, detector/VCO gain 0.3) instead of hard-coded constants. Output on real captures is unchanged (re-verified 100% symbol agreement against the updated C++ on three captures at 40 MHz and 62.5 MHz); the faster loop tracks wow/flutter substantially better per the reference implementation's testing.

PR #1058, opened 2026-07-31
Restore some decode throughput lost since April

JustinTArthur:
### Checklist
- [X] I have searched the open pull requests to confirm this change has not already been submitted.
- [X] My branch is up to date with the target branch.
- [X] I have tested my changes and all existing tests pass.
- [X] I have updated documentation where necessary.
- [X] My code follows the project's coding standards (see [CONTRIBUTING.md](CONTRIBUTING.md)).

### Description
Four independent decoder/demodulator performance regressions roughly halved throughput (including downstream in vhs-decode) and this tries to bring a little performance back.

1. `LoadFFmpeg`/`LoadLDF` rewind buffer (biggest one). Appending to and trimming the rewind buffer copies the bytes twice each read. 8a42fec2 raised `LoadFFmpeg`'s window from 2MB to 16MB, which turned ~4MB of memcpy per 61KB read into ~32MB. Demodulator threads starve waiting on the big copy.

2. `scale_field` lost `@njit(cache=True)`. Was dropped in 163408c5, I'm guessing an artifact of runtime kernel overrides while testing the change.

3. Sinc LUT's inter-phase interpolation doesn't change anything. `sinc_phase_count` is 2**16, so the nearest tabulated phase is already accurate below float32 precision. Interpolating between the phases is costly when this math is encountered frequently.

4. Wasteful `np.concatenate` on block record arrays. `nb_concatenate` was dropped in 980f3a2d and cf2e13ed but the replacement had an expensive deep traversal of block structure.

### Motivation
When vhs-decode 0.4.0 launched with these regressions, users brought up the slow performance in the Domesday86 Discord.

### Changes Made
1. Fixed with a `bytearray` and compaction only once the window has grown to twice its target size. The combination of those two is was better than just either one.
2. Added the `cache=True` back. @eshaz: let me know if you'd wanted to make tap counts or something else runtime-alterable.
3. Dropped the linear interpolation to just publish the nearest phase. Depends on the LUT staying this large.
4. Custom concatenation function. @happycube I asked an AI to determine why the numba flow had to be dropped, and it hinted you might have run into `can't unbox heterogeneous list`; this _should_ be faster than the numba kernel anyway.

### Testing
- [X] All existing tests pass (`pytest --output-on-failure`)
- [X] Tested manually with a barrage of LaserDisc (and VHS w/ patched vhs-decode) sources in both NTSC and PAL. TBC output was identical.

### DIFF
--- a/lddecode/core.py
+++ b/lddecode/core.py
@@ -34,6 +34,7 @@ from .utils import LRUupdate, clb_findbursts, angular_mean_helper, phase_distanc
 from .utils import build_hilbert, unwrap_hilbert, emphasis_iir, filtfft
 from .utils import fft_do_slice, fft_determine_slices, StridedCollector, hz_to_output_array
 from .utils import Pulse, nb_std, n_ornotrange, gen_bpf_supergauss, FieldInfo
+from .utils import concatenate_blocks
 
 try:
     # If Anaconda's numpy is installed, mkl will use all threads for fft etc
@@ -1400,7 +1401,7 @@ class DemodCache:
 
         rv = {}
         for k in t.keys():
-            rv[k] = np.concatenate(t[k]) if len(t[k]) else None
+            rv[k] = concatenate_blocks(t[k]) if len(t[k]) else None
 
         if rv["audio"] is not None:
             rv["audio_phase1"] = rv["audio"]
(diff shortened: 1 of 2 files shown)


eshaz (approved):
Tested with vhs-decode and indeed the performance is much improved. Thanks!

PR #1066, opened 2026-09-04
Decode performance: 1.4-2.0x throughput in less memory, by deleting work from hot paths

simoninns:
### Checklist

- [x] I have searched the open pull requests to confirm this change has not already been submitted.
- [x] My branch is up to date with the target branch.
- [x] I have tested my changes and all existing tests pass.
- [x] I have updated documentation where necessary.
- [x] My code follows the project's coding standards (see [CONTRIBUTING.md](CONTRIBUTING.md)).

### Description

Nine commits of decode throughput work sitting directly on `chad-2026.07.04-prev8-parallel`. On the
reference box the decoder is **2.0× faster on PAL CVBS, 1.6× on PAL `--tbc` and 1.4× on NTSC CVBS**
at their best thread counts, and **1.5–1.8× serially** — in less resident memory in every cell of
the grid, with no conformance band widened and `-t 1`/`-t N` byte-identity intact throughout.

Reference box: Ryzen 7 5800X (8C/16T, 32 MiB L3, dual-channel DDR4). PAL material is a Domesday CAV
capture, NTSC is Bambi, both `-s 5000 -l 1000`. Post-setup fps as the decoder reports it; RSS is the
peak across the whole process tree.

| Cell | `-t 1` | `-t 2` | `-t 4` | `-t 6` | `-t 8` |
|---|---:|---:|---:|---:|---:|
| PAL CVBS, before | 2.71 | 4.67 | 4.66 | 4.79 | **4.81** |
| PAL CVBS, **after** | 4.00 | 8.09 | 9.72 | **9.79** | 9.64 |
| PAL `--tbc`, before | 2.82 | 5.43 | **7.38** | 6.99 | 6.69 |
| PAL `--tbc`, **after** | 4.33 | 8.25 | **11.77** | 11.07 | 10.03 |
| NTSC CVBS, before | 3.01 | 5.56 | 9.15 | 10.01 | **10.50** |
| NTSC CVBS, **after** | 5.47 | 9.54 | **14.62** | 14.53 | 13.29 |

Peak tree RSS falls in every cell — PAL CVBS `-t 4` 2503 → 2129 MB, PAL `--tbc` `-t 8` 3836 → 3323,
NTSC CVBS `-t 1` 751 → 618.

Two shape changes matter as much as the levels. **PAL CVBS is no longer flat from `-t 2`** — it used
to sit at 4.67–4.81 across `-t 2`…`-t 8`, the signature that was read at the time as a
parent-process ceiling, and it now doubles from `-t 2` to `-t 4` before flattening. And **NTSC CVBS
stopped scaling to `-t 8`**: nothing about it got worse, its serial rate rose 82%, so four workers
now stream what eight used to and the knee moved down to meet them.

The full record — what was measured, what landed, what was built and reverted, and what was priced
and declined — is in [`plans/decode-performance.md`](plans/decode-performance.md).

### Motivation

The decoder stopped converting cores into frames at about four, and PAL CVBS stopped converting them
at two. The work set out to find why and move it.

The starting hypothesis was footprint: one decoder keeps ~11.5 MiB hot, two fit a 32 MiB L3 and three
do not, which is exactly where the concurrent-serial curve turned over. **That was tested directly
and it is false.** Rebuilding the sinc resample LUT took the hot set from 11.53 to 7.55 MiB — L3
would now hold 4.2 decoders where it held 2.8 — and throughput went **down 1–3% in every cell,
including at four concurrent decoders where the argument was supposed to apply.** The counters say
why: the 1.1e6 line fetches per frame vanished exactly as designed, and they were being subtracted
from **L3, not DRAM**. Nothing was waiting on them, and the interpolation the small table needs is
real arithmetic that nothing hides.

That result redirected everything after it. A change that removes cache lines pays only if those
lines were on a stalled path, and the counters can say whether they are *before* the change is made.
What replaced the premise is bytes *streamed* per frame, which is a different quantity from bytes
resident and behaves differently under contention.

### Related Issues

Related to #1061 and #1062, which optimise the sync demodulation FFT path. Those target `main` and
edit `lddecode/core.py`; this branch is on the prev8 line where that logic lives in
`lddecode/rfdecode.py`, so they do not conflict, but they are adjacent work and a reviewer may want
to look at them together.

### Changes Made

Each row measured on the committed harness against the tree immediately before it, interleaved and
plain:

- **`cache=True` on `scale_positions`**, matching `scale_field` (`dsp.py`) — a 1.99 s JIT compile
  becomes a 4 ms cache load, taking **1.83 s off every process that resamples**.
- **Hold `Filters["MTF"] ** mtf_level` per adopted level** instead of raising it per block
  (`rfdecode.py`) — block 3.017 → 2.721 ms on PAL; NTSC's `MTF` is `complex128` so its `pow` was
  2.29 of 5.18 ms, giving block −46% and **NTSC CVBS `-t 1` 3.01 → 4.83 fps (+60%)**.
- **Two further block hoists** (`rfdecode.py`) — PAL carrier notch applied in place; video and audio
  channels cast on assignment instead of into copies. Three 512 KiB temporaries deleted.
- **Chroma-DG correction transformed at `next_fast_len`** (`field.py`) — PAL CVBS field B is 354,689
  samples, which is prime, so scipy fell back to a generic radix: **105.6 → 9.1 ms**, worth
  **+18.7% (`-t 1`) to +43.5% (N=4)**. Padding tiles the field's own periodic continuation rather
  than zero-padding, so the circular convolution keeps the neighbour it already had.
- **256-phase interpolated sinc LUT** (`dsp.py`, `sinc_lut.npz`) — 4.00 MiB → 16 KiB. Kept on its own
  merits, not for throughput (see Motivation): both resample kernels now read the table identically
  where they previously did not, and the shipped table is *more* accurate than the nearest-neighbour
  lookup it replaced (7.8e-7 rms against a 3.05e-5 16-bit LSB). Its build also had a latent bug — the
  guard row was a copy of its neighbour rather than the phase-1.0 filter, invisible at 65536 phases
  and twenty times worse than the table it replaced at 256.
- **Chroma-DG corrector rebuilt** (`field.py`, `dsp.py`) — single precision through the filtering,
  the analytic chroma as a band with its quadrature via two *real* transforms, and one fused `njit`
  elementwise pass with no temporaries. Per field **21.6 → 6.1 ms** (gain), 27.2 → 8.9 (gain and
  phase), and 42.9 → 16.3 contended at `-t 6`. **PAL +10–19%, NTSC +2%.**
- **PAL CVBS burst lock damped to `g = 0.75`** (`cvbs.py`) — quality, not throughput: residual rms
  falls 8–11% on all ten PAL radius cuts and on the reference capture. This is the one output change
  in the PR; see Additional Notes for how it was found.
- **EFM output as one real inverse transform** of the folded half spectrum, rather than a complex one
  whose imaginary half is discarded (`rfdecode.py`) — per block **0.314 → 0.166 ms**, 9428 → 8043 L2
  misses, **bit-identical** output over 2.6M samples.
- **Video product stack filtered in float32** (`rfdecode.py`), with blanking subtracted before the
  cast and each channel's DC gain restoring it — per block **0.583 → 0.451 ms**, L2 misses 4.93e4 →
  2.86e4, **PAL `--tbc` `-t 6` +12.4%**. float32 carries 24 bits, so at the 8.5 MHz demod carrier its
  quantum is ~1 Hz — which is already the quantum these channels are *stored* at, since the record
  array they go into is float32 and always has been. Centring the block on blanking before the cast
  costs 0.044 ms of the 0.176 it protects and buys a factor of five: worst error 0.5 Hz = 0.00006 IRE,
  one storage step.
- **`fileio.SampleRing`** — one fixed allocation with a write and a read cursor, replacing a
  `bytearray` that re-seated its 64 MB buffer on nearly every extend. Reader thread **56.6 → 32.6
  ms/frame**; **PAL CVBS +8.8% (`-t 1`), +7.4% (`-t 6`), PAL `--tbc` +5.2%**. Bit-identical, 91 lines.

Supporting work:

- Three committed measurement harnesses: `scripts/bench_decode_throughput.py` (one
  `(system, mode, -t, -s, -l, capture)` cell as post-setup fps and peak tree RSS),
  `scripts/report_working_set.py` (resident filters, LUT, bytes `demodblock` indexes per block,
  block temporaries under `tracemalloc`) and `scripts/report_decode_traffic.py` (hardware counters
  attributed per named decoder stage). `scripts/build_sinc_lut.py` regenerates the shipped table.
- `AGENTS.md` §6 now points at the first two, so figures from different changes stay comparable.
- Seven new unit suites covering the block hoists, the DG corrector, the sample ring, the sinc LUT,
  the traffic instrument and the harness helpers.

### Testing

- [x] The unit lane passes: `python -m pytest -q tests/unit` — **1754 passed, 3 skipped**
- [x] The functional lane passes, or is untouched: `ctest --test-dir build --output-on-failure` —
      **95/95 functional, 0 failed**; the two not run are `roundtrip-{ntsc,pal}-orc`, which take the
      documented skip path when `orc-cli` is absent
- [x] Unit tests added or updated in the same PR as the behaviour change.
- [x] Every new test is marked `unit` or `functional`, and labelled to match in CMake where registered.
- [x] Unit tests touch no filesystem, network, subprocess or clock.
- [x] Every generator is seeded; every float assertion carries a stated tolerance.
- [x] The layer boundaries in [AGENTS.md](AGENTS.md) §2 still hold.
- [x] For decode changes: the serial/threaded comparisons and the CVBS verifier still pass —
      `compare-{ntsc,pal,jason-pll}-parallel-{tbc,pcm,efm}` and
      `compare-pal-cvbs-parallel-{cvbs,meta,efm,audio}` all pass, and
      `analysis/vits_known_deviations.toml` is untouched from the first commit to the last.
- [x] For format changes: N/A — no output format changed.
- [x] Any intentional skip is documented in the test body with a reason.
- [x] Tested manually with: the throughput grid above (three cells × five thread counts, before and
      after, interleaved in one session on an idle box), plus N independent serial decoders at
      N = 1/2/4/8, and per-stage counter attribution on PAL CVBS and PAL `--tbc`.

Two test consequences are called out because they are the kind that get papered over.
`demodblock_sync` is held to *bit* equality with `demodblock`'s sync channel; rather than loosening
that assertion it was moved to the same precision and the same stack row, so it still passes on
`assert_array_equal`. And `tests/unit/test_demod_fft.py` compared the pipeline against the exact
double-precision answer at `rtol=1e-6`, which no longer describes what the pipeline promises; rather
than widen it, the assertion was restated in the units that matter — within four float32 steps of
each channel's peak, which is the error growth a 32768-point transform pair is entitled to. It fails
on a regression instead of absorbing one.

### Additional Notes

**Three changes were implemented in full, proved correct, measured, and taken back out.** Together
they are the most transferable result here. The EFM demodulator on its own output lane is correct and
did what it was meant to (`ld-output` 79.2 → 68.5 ms/frame) but buys nothing, because the parent was
already thread-saturated and the new thread made every other thread ~30% more expensive — 55 ms/frame
of new work to move 28 ms off one thread. The PAL 4fsc resample in the workers shed what it was meant
to shed from the parent (178% → 123% of a core at `-t 4`) but the workers gained far more than the
parent lost (333% → 511% at `-t 6`). And the VITS measurements moved into the workers bit-identically
and cost 2.0% on the binding cell.

> Work is not movable between these ceilings at par. The only changes that have paid are the ones that
> **delete** work. Moving it — to another parent thread, or into the workers — costs about 3× what it
> saves.

The cleanest data point is the `SampleRing` change against the VITS revert: both are parent-side items
of the same kind and order of size (24 and 7 ms of parent CPU per frame), both bit-identical, both
measured interleaved in the same session on the same box. The one that *deleted* the work paid
**+7.4%**; the one that *moved* it cost **−2.0%**.

**The resample experiment exposed a real fault worth more than the change.** The PAL burst lock was
chasing its own measurement noise: at a tolerance below the burst measurement's noise, 99.5% of
shipped resamples were redone, and over 120 frames the residual has a mean of 0.01° against an rms of
1.65 — a bounded random walk, not a drift. Applying the whole of each frame's residual writes the
previous frame's noise into the next frame's lattice. Damping the loop is the `g = 0.75` change above.

**Known limitation — the `-t` auto default is above the measured optimum.** `lddecode/main.py`
computes `min(max(physical_cpu_count() - 2, 1), 10)`, which is 6 on the reference box; the optimum
measured here is 4 in every cell. At `-t 6` PAL CVBS and NTSC are unchanged within the run-to-run
spread, but **PAL `--tbc` loses 6.3%** and the default costs **630 MB of peak tree RSS**. It is
deliberately *not* changed in this PR: the knee is a property of the box, and re-deriving the default
wants a unit test over injected core and memory counts rather than this box's number hard-coded.

**Where the decoder stands.** The concurrency architecture is not what limits it. PAL `--tbc` at
`-t 4` delivers 11.77 fps and four *independent* serial decoders deliver 11.74 — the two best PAL
figures the box produces by any arrangement, agreeing to 0.3% — and eight independent decoders return
*less* than four do. What binds past N ≈ 4 is DRAM traffic manufactured by concurrency itself, at
7.6 GB/s against a path measured at 20 GB/s: the same instructions per frame to within 4% across the
whole curve, for 1.22× the cycles at N = 4 and 2.44× at N = 8. The remaining lever is bytes streamed
per frame inside the demodulator, where the block chain is 81% of a PAL CVBS frame's L2-miss traffic.
§8 of the document lists what is left, in order of what the evidence supports.

**Every figure is one box** — one 5800X, one L3 size, one DRAM speed. The knees are stated, not
assumed to transfer, and the harness records the box.


simoninns:
I've added the MTF fixed dp11 spotted into the PR (this is sitting at the top of my branches - so I have to commit up to keep track) - feel free to merge asap ;)

PR #1067, opened 2026-09-05
FFT Optimisation

simoninns:
### Checklist

- [x] I have searched the open pull requests to confirm this change has not already been submitted.
- [x] My branch is up to date with the target branch.
- [x] I have tested my changes and all existing tests pass.
- [x] I have updated documentation where necessary.
- [x] My code follows the project's coding standards (see [CONTRIBUTING.md](CONTRIBUTING.md)).

### Description

Four commits on top of `chad-2026.07.04-prev8-parallel`, in two unrelated groups.

**1. The FFT/demodulator round** (`4dc63491`, plus the conformance follow-up in `61916fac`).
Moves the FM discriminator onto a half-rate analytic signal, deletes work from the block chain,
reuses demodulated blocks across adjacent field jobs, and shares one filter-bank mapping per
decode instead of one bank per worker. Serial decoding is **7.7% faster on PAL CVBS, 13.5% on
PAL `--tbc`, 14.8% on NTSC CVBS**; the demodulator's hot working set per block falls from
7.36 to **5.74 MiB** on PAL, which is worth nothing to a decoder running alone and **10-11% of
every frame's DRAM fills among four or eight concurrent decoders**. The full record — including
what was built and reverted, what was priced and declined, and the one measurement that would
close the round — is in `plans/fft-optimisation.md`.

That work also surfaced and fixed a pre-existing calibration defect: the 2T servo was acting on a
stale inverse-MTF bound. Fixing it took `domesday-ds2-community-north-middle` from 7 real VITS
failures to 0, and three now-passing entries were deleted from
`analysis/vits_known_deviations.toml`, as that file's contract requires.

**2. A resampler coordinate-precision fix** (`3e5f7359`, plus `c02a4ecd`). Reported by **dp11**.
Both resample kernels in `lddecode/dsp.py` narrowed the source sample position to float32 before
splitting off its fractional part. A field's demod buffer runs to roughly 8e5 samples, where
float32 spacing is 1/32 of a sample from 2^18 and 1/16 from 2^19 — so the lower two-thirds of
every field was snapped onto a coarse grid and the phase resolution the interpolated 257-row sinc
table exists to provide was discarded. Measured on a band-limited 6.3 MHz signal at the PAL input
rate, the quantisation added broadband error at **-45 dB** relative to the signal in the middle
third of a field and **-40 dB** in the bottom third. Keeping the coordinate in float64 until the
integer part is split off removes it at no cost: the tap loop and `phase_alpha` stay float32, and
`scale_field` on a full PAL field measures 6.3 ms after against 6.5 ms before.

### Motivation

The performance record (`plans/decode-performance.md` §8) left "bytes per frame inside the
demodulator" as the next item the evidence supported, with a precondition attached; that
precondition is met and the item is closed. The premise it confirms is that on this decoder,
footprint claims have to be measured among N concurrent decoders — resident size predicts nothing,
and a change that frees megabytes from a worker's hot set buys nothing solo and double digits once
L3 is genuinely contended.

The resampler fix is unrelated and independent: it is a straightforward loss of precision that
cost the resampler most of its sub-sample accuracy over the majority of every field, in the part
of the pipeline whose entire job is to place samples accurately in time.

### Related Issues

None. Two open PRs touch adjacent ground but target `main`, not this branch, and neither overlaps
these changes: happycube/ld-decode#1061 "Reduce sync demodulation FFT work" and the draft
happycube/ld-decode#1062 "Optimize sync FFT path and refresh benchmark filters".

### Changes Made

- `4dc63491` — **Optimise FFT implementation.** Half-rate analytic discriminator with the
  `exp(+i·pi·f·T)/cos(pi·f·T)` rate correction; byte-identical deletions in the block chain
  (half-spectrum filtering, conjugate-symmetry audio slice, thread-local scratch, five
  construction-only filters retired); block reuse across adjacent field jobs (`AffinityPool`,
  `WorkerBlockLRU`); a shared-memory filter segment (`lddecode/shared_filter_bank.py`); developer
  controls (`LDDECODE_BLOCKLEN`, `LDDECODE_BLOCK_LRU_STATS`, `LDDECODE_SHARED_FILTER_STATS`,
  `LDDECODE_NO_SHARED_FILTERS`) documented in `docs/user-guide/scripts.md`; and the 2T servo
  stale-bound fix. Adds ~2,900 lines of unit tests across nine new or extended files.
- `61916fac` — **Fix a VITS conformance issue.** Deletes the two
  `domesday-ds2-community-north-outer` entries the run now passes, which the deviation list's
  contract requires be deleted rather than left to rot.
- `3e5f7359` — **Fix the resampler position quantisation.** Two lines in `lddecode/dsp.py`, plus
  a regression test that resamples the same fractional positions near the start of a buffer and
  2^20 samples in and requires them to agree; it fails without the fix by 0.088 on a unit-rms
  signal, across 99.6% of positions. The test file's own `resample_nearest` reference stopped
  narrowing to float32 in the same commit, so it no longer carries the error it is meant to
  isolate.
- `c02a4ecd` — **Record one VITS check as a known deviation.** See "Additional Notes".

### Testing

- [x] The unit lane passes: `python -m pytest -q tests/unit` — **2015 passed, 3 skipped**.
- [x] The functional lane passes: **48 of 48** in CI on `3e5f7359` for the functional job, with
      the VITS lane's one failure addressed by `c02a4ecd` and re-verified locally.
- [x] Unit tests added or updated in the same PR as the behaviour change.
- [x] Every new test is marked `unit` or `functional`, and labelled to match in CMake where registered.
- [x] Unit tests touch no filesystem, network, subprocess or clock.
- [x] Every generator is seeded; every float assertion carries a stated tolerance.
- [x] The layer boundaries in [AGENTS.md](AGENTS.md) §2 still hold.
- [x] For decode changes: the serial/threaded comparisons and the CVBS verifier still pass —
      every `compare-*-parallel-*` test passes on the final tree, so the `-t N` bit-identity
      guarantee in AGENTS.md §4.4 holds.
- [x] For format changes: N/A, no format changed.
- [x] Any intentional skip is documented in the test body with a reason.
- [x] Tested manually with: `dolby-surround-side1-outer` decoded twice — current tree and with
      `lddecode/dsp.py` from `4dc63491` — and both run through
      `analysis/vits_conformance.py`, to attribute the one moved check.

### Screenshots (if applicable)

N/A.

### Additional Notes

**The resampler fix changes output bytes for every decode.** That is the point of it, but it means
any recorded reference decode is invalidated. On the NTSC cut measured here, 33 of 50 conformance
checks moved and every one of them moved by **0.031 IRE or less** — third-decimal changes, no
amplitude metric materially affected.

**One VITS check crossed its limit as a result, and `c02a4ecd` records it.**
`ntsc-ntc7-composite/pulse_12t5_luma` on `dolby-surround-side1-outer` went from 48.512 to
48.481 IRE against a nominal of 50.000 ±1.500 — from 0.99x of its band to 1.01x. The check was
already spending 99.2% of its allowance before this PR, and the same check on
`dolby-surround-side1-middle` is already listed at 1.24x, so this is the third radius of a fault
the deviation list already tracks at two, not something the resampler introduced. It is recorded
at the run's own figure with the customary 1.10x ceiling, and its `reason` deliberately does not
reuse the sibling entries' wording: the 2T pulse grows outwards and this check does not.

**One inconsistency a reviewer will notice.** `plans/fft-optimisation.md` §4.4 states that the
VITS radius lane has "no real failures on any cut" and that the whole lane is green at 96 of 96.
That was true when it was written and is true of the FFT round on its own, but `c02a4ecd` adds an
entry back. The claim in that section should be amended to say so before merge, or the section
read as scoped to the FFT work.

**On the branch's base.** This branch does not carry upstream's merge commit `edb8a9b0`
(the merge of #1066), but that merge introduced no content: its tree is identical to the merge
base `6a15fa29`, so the four commits here apply to the current target with no rebase and no
conflict.

**What could not be measured.** The PAL cells at `-t 4` and `-t 6` are 30-frame decodes — nothing
in `testdata/` is longer than 90 frames and the long reference captures live on a NAS mount that is
empty on the measuring box. They never reach steady state and show no difference, where NTSC, on
the only capture long enough, keeps +17.7% and +15.8%. A long PAL capture at `-t 4` is the one
measurement that would close the round, and it would settle "the gain is serial-only on PAL"
against "the cell was too short to see it".


simoninns:
Hold off on merging this for a bit - I need to investigate something odd showing up in decode-orc

simoninns:
Ok merging is fine - here is the reasoning behind the SNR analysis shift:

## PAL black PSNR: ~5 dB lower than 2025 decodes (expected)

Comparing a DS2 NationalA decode from this branch against the January 2025
decode of the same disc shows the black PSNR trace ~5 dB lower. This is the
documented cost of the PAL video LPF widening, not a regression.

### Measurements

Domesday DS2 NationalA, frames 5000–25000, field 1, measured on
decode-orc's own slices (PAL line 22 @ 12–62 µs black, line 19 @ 12–20 µs white):

| | 2025 decode | this branch | Δ |
|---|---|---|---|
| black line RMS | 1.337 IRE → 37.5 dB | 2.307 IRE → 32.7 dB | **−4.7 dB** |
| white bar RMS | 3.170 IRE → 30.0 dB | 3.272 IRE → 29.7 dB | −0.3 dB |

The extra energy is a band at 5.4–6.8 MHz, +6 to +12 dB; below 5 MHz the two
decodes agree within 1–2 dB. Re-filtering this branch's output with the old
5.2 MHz order-7 Butterworth restores 37.03 dB — within 0.45 dB of the 2025
decode — so the entire difference is the LPF.

### Cause

The PAL video low-pass corner moved in two steps:

| | `video_lpf_freq` | `video_lpf_order` | black PSNR |
|---|---|---|---|
| Jan 2025 baseline | 5.2 MHz | 7 | 37.5 dB |
| f7ddee5d (2026-05-30) | 5.8 MHz | 7 | 34.6 dB |
| 36939ea3 (2026-09-02) | 6.3 MHz | 16 | 32.7 dB |

36939ea3 predicted 6.3/16 would cost 1.23× the FM-weighted noise of 5.8/7;
measured 1.869 → 2.307 IRE = 1.234×. The filter behaves exactly as designed.
The baseline in that commit's analysis was 5.8 MHz, so against a 2025-era
image the cumulative cost is 1.73× RMS rather than 1.23×.

Demodulated FM noise power rises as f², and the passband now extends ~0.8 MHz
past the 5.5 MHz System I luma band, so that last stretch carries far more
noise than signal. This is the deliberate trade for containing the 5.8/5.9 MHz
VITS multiburst packet inside the passband instead of on its −3 dB corner.

